PR #13, opened 2018-08-15
Remove info graphs

alegnaaived:
I moved all the charts to one folder outside modules folders, and it's necessary for new fixes to check eslint rules in others charts (BarGraph and DotsGraph).


src/charts/BarStackHorizontal.jsx
```diff
@@ -30,38 +31,42 @@ export default withTooltip(
...
+      });
       return transformedData;
-    }
+    };
 
     function comparingValues(a, b) {
       return a - b;
     }
 
+    let data = [prepareData(dataJSON, labelY)].slice(0);
```
erikasv: This is unnecessary, data should be just `prepareData(dataJSON, labelY)`, and so, the next line would be just `Object.keys(data)`

src/charts/BarStackHorizontal.jsx
```diff
@@ -111,44 +119,45 @@ export default withTooltip(
               xScale={xScale}
               yScale={yScale}
               zScale={zScale}
-              onClick={data => event => {
-                if (!events) return;
-                alert(`clicked: ${JSON.stringify(data)}`);
-                // actualizarSubArea(data.key);
-              }}
-              onMouseLeave={data => event => {
+              // onClick={data => (event) => {
```
erikasv: you should remove unused code

src/charts/BarStackHorizontal.jsx
```diff
@@ -14,7 +15,7 @@ export default withTooltip(
     width,
     height,
     graphTitle,
-    events = false,
+    // events = false,
```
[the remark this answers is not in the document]
alegnaaived: Solved

src/charts/BarStackHorizontal.jsx
alegnaaived: Solved

src/charts/BarStackHorizontal.jsx
alegnaaived: After pair review, it is now solved

PR #16, opened 2018-08-17
Compensations adjustments

erikasv:
* Creation of TabContainer component to render tabbed content
* Adjust to listeners and actions on "donde - como" section on compensations module


src/compensator/Drawer.jsx
```diff
@@ -175,11 +166,15 @@ class Drawer extends React.Component {
...
     });
-    this.setState({
+    this.setState(prevState => ({
       szh,
       car,
       strategies,
-    });
+      showGraphs: {
+        ...prevState.showGraphs,
+        graphDonde: false,
```
alegnaaived: ¿graphWhere or something different?

alegnaaived: I'm considering as a strategy, removing all Spanglish mixes on the code while we are refactoring. What do you think?

erikasv: `DotsWhere` ?

erikasv: I agree, I just didn't know if we should use common words like "What" anda "Where", but I think it's the same as in Spanish.

alegnaaived: Definitely, maybe the best way is adding something like "DotsWhereGraph"

erikasv: but it's too long, besides, since it's inside an object called "showGraphs" it's obvious we are talking about graphs, that's why I was thinking to use like a pettern <GraphType><SectionName>. So, if we end up with BarStackHorizontalCompensation we don't make it longer with the "Graph"

alegnaaived: Okay, I agree

PR #30, opened 2018-08-30
Colorizing graphs

alegnaaived:
Creating structure for colors, sending colors from Search and removing colors from each chart, leaving one as default: blue. Also, removing restriction on geoserver api.


src/Search.jsx
```diff
@@ -100,59 +139,64 @@ class Search extends Component {
...
+    point.setStyle({
       weight: 1,
       fillOpacity: 1,
     });
     switch (parentLayer) {
       case 'jurisdicciones':
-        event.target.bindPopup(event.target.feature.properties.IDCAR);
+        point.bindPopup(
+          `<b>${point.feature.properties.IDCAR}</b>
+          <br>${point.feature.properties.NOMCAR}`,
```
erikasv: don't break this line https://github.com/airbnb/javascript#whitespace--max-len

src/Search.jsx
```diff
@@ -100,59 +139,64 @@ class Search extends Component {
...
+        );
         break;
       case 'corpoBoyaca':
-        event.target.bindPopup(
-          `Bioma: ${event.target.feature.properties.BIOMA_IAvH}<br>Factor de compensación: ${event.target.feature.properties.FC_Valor}`,
+        point.bindPopup(
+          `Bioma: ${point.feature.properties.BIOMA_IAvH}<br>Factor de compensación: ${point.feature.properties.FC_Valor}`,
```
erikasv: don't break this line

src/Search.jsx
```diff
@@ -100,59 +139,64 @@ class Search extends Component {
...
-    ElasticAPI.requestBiomaBySZH(bioma)
-      .then((res) => {
-        this.setState(prevState => ({
-          geojsonCapa4: bioma,
-          activeLayers: {
-            ...prevState.activeLayers,
+    switch (parentLayer) {
+      case 'corpoBoyaca': {
+        const bioma = event.target.feature.properties.BIOMA_IAvH;
+        ElasticAPI.requestBiomaBySZH(bioma)
```
erikasv: It's better to let this call in a separate function, in the future it will be called too when click on the bioma from the graph, right?

src/charts/BarStackHorizontal.jsx
```diff
@@ -165,8 +154,7 @@ export default withTooltip(
               </strong>
             </div>
             <div>
-              {Number(tooltipData.data[tooltipData.key]).toFixed(2)}
-                Ha
+              {`${Number(tooltipData.data[tooltipData.key]).toFixed(2)} Ha`}
```
erikasv: Very good catch :)

src/Search.jsx
alegnaaived: Agree. I created this:
// TODO: Create function for jurisdicciones layer

PR #31, opened 2018-08-30
Fix compensation

erikasv:
* Add missing label in que-cuanto, donde-como section
* Adjust index.js to pass eslint vaidations
* Change button behavior when entered area exceeds limit in donde-como in compensations.

### DIFF
--- a/src/compensation/Drawer.jsx
+++ b/src/compensation/Drawer.jsx
@@ -108,19 +108,20 @@ class Drawer extends React.Component {
    * @param {number} maxValue maximum allowed value
    */
   operateArea = (value, operator) => {
-    switch (operator) {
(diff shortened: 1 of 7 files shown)


src/compensation/Drawer.jsx
```diff
@@ -108,19 +108,20 @@ class Drawer extends React.Component {
...
-          tableError: '',
-        }));
-        break;
-      case '-':
-        this.setState(prevState => ({ selectedArea: prevState.selectedArea - value }));
-        break;
-      default:
-        break;
-    }
+    this.setState((prevState) => {
```
alegnaaived: Better practices! Our code is looking better each time :dancer: :dancing_women:

PR #36, opened 2018-08-30
Fix map layers

erikasv:


src/Compensation.jsx
```diff
@@ -30,15 +29,14 @@ class Compensation extends Component {
       GeoServerAPI.requestSogamoso(),
       GeoServerAPI.requestBiomasSogamoso(),
```
jsricarde: Move the fetch data functions to componentDidMount or componentWillMount mehtods https://daveceddia.com/where-fetch-data-componentwillmount-vs-componentdidmount/

PR #50, opened 2018-09-06
Fixing interactions

alegnaaived:
Interactions between DotsGraph and the map


src/Compensation.jsx
```diff
@@ -124,15 +171,17 @@ class Compensation extends Component {
     if (!L.Browser.ie && !L.Browser.opera) area.bringToFront();
   }
 
-  resetHighlight = (event, layer) => {
-    const feature = event.target;
-    const { layers } = this.state;
-    layers[layer].layer.resetStyle(feature);
+  resetHighlight = (area, parentLayer) => {
+    const { layerName, layers } = this.state;
+    if (layers[parentLayer] && layerName
```
erikasv: You should change the structure of this if: https://github.com/airbnb/javascript#control-statements
Also I think it's better to use braces because of the long condition

PR #59, opened 2018-09-11
Config file

alegnaaived:
Creating config files and editing routes and paths, setting user object


package.json
```diff
@@ -25,6 +25,7 @@
     "eslint": "4.19.1",
     "leaflet": "1.3.3",
     "prop-types": "15.6.2",
+    "topojson": "3.0.2",
```
erikasv: remove this dependency and include it when you're going to use it

alegnaaived: Done

PR #104, opened 2018-11-22
Compensaciones: Add new status option when creating a Project

alegnaaived:
We create the option for adding new states in projects

### DIFF
--- a/src/compensation/NewProjectForm.jsx
+++ b/src/compensation/NewProjectForm.jsx
@@ -12,6 +12,7 @@ class NewProjectForm extends Component {
       regionSelected: null,
       statusSelected: null,
       newName: null,
+      newNameState: null,
     };
   }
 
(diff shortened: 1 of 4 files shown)


erikasv:
testing issue linking: #103

src/compensation/NewProjectForm.jsx
```diff
@@ -57,15 +58,29 @@ class NewProjectForm extends Component {
...
-      />
+      <div>
+        <Select
+          value={statusSelected}
+          onChange={this.handleChangeStatus}
+          placeholder="Estado del proyecto"
+          options={status}
+        />
+        { // TODO: Handle error for new project if the company doesn' have regions and status
+          (statusSelected === status[status.length - 1].value) && (<br />) && (
```
erikasv: I think you should hardcode 'newState', instead of "status[status.length - 1].value", the later isn't clear (and could break if someone adds a new item to the list in the future)

alegnaaived: Agree 
```suggestion
          (statusSelected === 'newState') && (<br />) && (
```

PR #143, opened 2019-03-19
Reordering graphs and fixing typos in hectares units

alegnaaived:


src/Search.jsx
```diff
@@ -209,7 +207,11 @@ class Search extends Component {
    * @param {String} parentLayer Parent layer ID
    */
   loadLayer = (idLayer, parentLayer) => {
-    this.setState({ loadingModal: true });
+    this.setState({
+      loadingModal: true,
+    });
+    // TODO: Implement new load strategy by geofence id
```
erikasv: I don't understand why is necessary to load strategies in this component, maybe add an explanation to this comment

alegnaaived: Not as a compensation strategy, but as a loading strategy. I changed the comment to avoid this interpretation.

PR #147, opened 2019-04-02
National Strategic Ecosystem

alegnaaived:
NationalInsigths implements three endpoints, included in RestAPI

### DIFF
--- a/src/Search.jsx
+++ b/src/Search.jsx
@@ -304,19 +304,19 @@ class Search extends Component {
   /** ****************************** */
   /** LISTENERS FOR SELECTOR CHANGES */
   /** ****************************** */
-  secondLevelChange = (name) => {
+  secondLevelChange = (id) => {
     const { areaList } = this.state;
     this.setState((prevState) => {
       let newState = { ...prevState };
       newState = {
         ...newState,
         area: areaList.find(
-          item => item.name === name,
+          item => item.id === id,
         ),
       };
       return newState;
     });
-    this.loadSecondLevelLayer(name);
+    this.loadSecondLevelLayer(id);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


src/Search.jsx
```diff
@@ -20,8 +21,8 @@ class Search extends Component {
     super(props);
     this.state = {
       activeLayers: null,
-      activeLayerName: null,
-      geofenceData: null,
+      activeLayer: null,
+      geofenceData: null, // TODO:
```
erikasv: Complete the TODO comment or delete it

src/Search.jsx
```diff
@@ -149,13 +150,14 @@ class Search extends Component {
   }
 
   highlightFeature = (event, parentLayer) => {
+    const { area } = this.state;
     const point = event.target;
     point.setStyle({
       weight: 1,
       fillOpacity: 1,
     });
     switch (parentLayer) {
```
erikasv: Review this switch statement, the "case area.id" means every parentLayer value will match that option, also the default wold never be used

src/Search.jsx
```diff
@@ -255,14 +255,15 @@ class Search extends Component {
    */
   loadSecondLevelLayer = (idLayer) => {
     const { areaList } = this.state;
+    // TODO: Change ot for a loading layer strategy
```
erikasv: I don't understand the comment, what is "ot"? and what do mean with "loading layer strategy"?

src/Search.jsx
erikasv: This case works the same as an if statement (parentLayer === area.id) and I think it's simpler to understand that way. Besides, the popup value only works for "jurisdicciones" objects (properties IDCAR and NOMCAR)

I think would be better the if statement and a TODO message for future layers

PR #170, opened 2019-05-10
Information per graph

alegnaaived:
Defining classes and components to show graphs information. Applied just to BarStackHorizontal

### DIFF
--- a/src/commons/ShortInfo.jsx
+++ b/src/commons/ShortInfo.jsx
@@ -1,5 +1,6 @@
 /** eslint verified */
 import React from 'react';
+import PropTypes from 'prop-types';
 
 class ShortInfo extends React.Component {
   constructor(props) {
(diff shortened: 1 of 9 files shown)


src/commons/ShortInfo.jsx
```diff
@@ -0,0 +1,74 @@
...
+    return (
+      <div>
+        <div
+          className={`${className}-${hideText}`}
+        >
+          <p>
+            <b>
+              {name}
+            </b>
+            {description && (` ${description}`)}
```
erikasv: The validation isn't necessary, just with {\` ${description}\`} is enough

src/commons/ShortInfo.jsx
```diff
@@ -0,0 +1,74 @@
...
+  name: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  description: PropTypes.string,
+  tooltip: PropTypes.string.isRequired,
+  customButton: PropTypes.bool,
+};
+
+ShortInfo.defaultProps = {
+  className: 'hidden',
+  description: NaN,
```
erikasv: Since it's an string its default value should be null or an empty string

src/commons/ShortInfo.jsx
alegnaaived: Fixed

src/commons/ShortInfo.jsx
alegnaaived: Fixed

PR #179, opened 2019-06-14
Details

alegnaaived:


src/charts/RenderGraph.jsx
```diff
@@ -14,14 +14,16 @@ import GraphLoader from './GraphLoader';
  * @param {string} graphTitle graph title
  * @param {array} colors colors to sort elements inside the graph
  */
+
```
erikasv: These blank lines shouldn't be there

PR #217, opened 2020-04-02
Accordion for landscapes tab

alegnaaived:
New component for this custom accordion.

Example 1 inside Landscape tab:
image

### DIFF
--- a/src/commons/Accordion.jsx
+++ b/src/commons/Accordion.jsx
@@ -1,0 +1,76 @@
+/** eslint verified */
+import React from 'react';
+import ExpansionPanel from '@material-ui/core/ExpansionPanel';
+import ExpansionPanelSummary from '@material-ui/core/ExpansionPanelSummary';
+import ExpansionPanelDetails from '@material-ui/core/ExpansionPanelDetails';
+import PropTypes from 'prop-types';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+
+import { dataPaisajes } from '../search/assets/selectorData';
+
+class Accordion extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      expanded: null,
+      selected: null,
+    };
+  }
+
+  firstLevelChange = panel => (event, expanded) => {
+    const { handlers } = this.props;
+    const expandedPanel = expanded ? panel : false;
+    handlers[0](expandedPanel);
+    this.setState(prevState => ({
+      expanded: expandedPanel,
+      selected: expanded ? panel : prevState.expanded,
+    }));
+    return null;
+  };
+
+  componentWillUnmounted() {
+    const { handlers } = this.props;
+    handlers[0](false);
+  }
+
+  render() {
+    const { children, description, iconClass } = this.props;
+    const { expanded, selected } = this.state;
+    const data = dataPaisajes;
+    return (
+      <div>
+        {(Object.keys(data).length > 0)
+          && data.map(counter => (
+            <ExpansionPanel
+              className={counter.className}
+              id={counter.id}
+              disabled={false}
+        // onChange={this.firstLevelChange(id)}
+              key={counter.id}
+            >
+              <ExpansionPanelSummary
+                expandIcon={<ExpandMoreIcon />}
+              >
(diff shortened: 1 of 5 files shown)


erikasv:
None of the commits is in the form we discuss *"if applied, this commit will…"* Please use the following command to change those messages: `git rebase -i 9dee58f5` (that's the commit from which you created the branch, so you can operate all commits in the branch)

Inside the editor, change all "pick" for "edit" (as it will say in the instructions) and follow the steps.

erikasv:
Ah!, after you make the rebase need to push your branch with "-f" option. After that I will check the code, if I do it before the "push -f" I'm not sure if the comments would persist.

src/search/CompensationFactor.jsx
```diff
@@ -0,0 +1,54 @@
...
+  compensationFactorColors,
+  compensationFactorData,
+  handlerInfoGraph,
+  openInfoGraph,
+) => (
+  <div className="graphcard">
+    <div className="graphcontainer">
+      <h4>
+      hectáreas totales
+        <b> 1,827,103.00 ha</b>
```
erikasv, line 20: There is an indentation error here, see eslint errors

src/search/CompensationFactor.jsx
```diff
@@ -0,0 +1,54 @@
+/** eslint verified */
+import React from 'react';
+import RenderGraph from '../charts/RenderGraph';
+
+const CompensationFactor = (
```
erikasv, line 5: This function is a React component, its only input must be an object (the props parameter), not the many parameters it has here.

Usually that object is destructed in the function signature, to avoid an extra sentence inside the function, but it still receives only one parameter.

You must fix these inputs and declare the propTypes of the component

src/search/Drawer.jsx
```diff
@@ -194,6 +198,41 @@ class Drawer extends React.Component {
...
+      component: CompensationFactor(
+        area.name, // areaName
+        colors, // biomesColors
+        biomas, // biomesData
+        colorsRB, // bioticRegionsColors,
+        distritos, // bioticRegionsData,
+        colorsFC, // compensationFactorColors,
+        fc, // compensationFactorData,
+        handlerInfoGraph,
+        openInfoGraph,
```
erikasv, line 219: Once the CompensationFactor component is fixed this will change and the comments won't be necessary (actually they are never needed, that's why functions have a signature)

src/search/CompensationFactor.jsx
cazapatamar: I could fix the component definition as i have to work on it for the task i have assigned.

erikasv: I prefer it to be fixed in this PR, the idea of the revisions are to avoid merging this kind of things, but we can discuss it in the daily

alegnaaived: As @erikasv said, we checked in the daily and it was accepted because this component should be modified in this sprint

src/search/Drawer.jsx
alegnaaived: This should be sent as several parameters until CompensationFactor being modified according to this sprint.

src/search/CompensationFactor.jsx
alegnaaived: Done

src/commons/Accordion.jsx
```diff
@@ -0,0 +1,101 @@
...
+        description: 'Representa diferentes análisis de huella humana en esta área de consulta',
+      },
+      component: RenderGraph(distritos, 'Hectáreas', 'Regiones Bióticas', 'BarStackGraph',
+        'Regiones Bióticas', colorsRB, handlerInfoGraph, openInfoGraph,
+        'muestra las hectáreas por cada región biótica en el área de consulta seleccionada'),
+    },
+    ];
+      */
+      classNameSelected, // className defined in CSS file to selected item this accordion
+      classNameDefault, // className defined in CSS file to default item for this accordion
```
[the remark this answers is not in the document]
alegnaaived: Done. It was moved to PropTypes definitions and changed the componentsArray definition as required and the data structure required.

PR #221, opened 2020-04-15
Component for Factor Compensation Graphs

cazapatamar:


src/search/CompensationFactor.jsx
```diff
@@ -1,45 +1,59 @@
...
+    compensationFactorColors,
+    compensationFactorData,
+    handlerInfoGraph,
+    openInfoGraph,
+  } = props;
+  return (
+    <div style={{ width: '100%' }}>
+      <div className="graphcontainer">
+        {(areaName && areaName === 'Jurisdicciones ambientales')
+          && RenderGraph(compensationFactorData, 'Hectáreas', 'F C', 'BarStackGraph', 'Factor de Compensación', compensationFactorColors, handlerInfoGraph, openInfoGraph, 'representa las hectáreas sobre los Biomas IAvH analizados')}
```
erikasv, line 22: Since this line is not only the string, it's better to break it to not longer than 100 characters, to make it more readable.

Also apply it to lines 24 and 26

src/search/Drawer.jsx
```diff
@@ -207,17 +207,17 @@ class Drawer extends React.Component {
...
       },
-      component: CompensationFactor(
-        area.name, // areaName
-        colors, // biomesColors
-        biomas, // biomesData
-        colorsRB, // bioticRegionsColors,
-        distritos, // bioticRegionsData,
-        colorsFC, // compensationFactorColors,
-        fc, // compensationFactorData,
+      component: CompensationFactor({
```
erikasv, line 210: Avoid calling components as pure functions. Better "render" the component <CompensationFactor>...

cazapatamar: Done

src/search/CompensationFactor.jsx
cazapatamar: Done

PR #225, opened 2020-04-17
Feature/refactor graphs components

cazapatamar:
Changes on RenderGraph and GraphLoader components to receive props

### DIFF
--- a/src/charts/GraphLoader.jsx
+++ b/src/charts/GraphLoader.jsx
@@ -12,27 +12,53 @@ import DotInfo from './DotInfo';
 
 ReactChartkick.addAdapter(Chart);
 
-const GraphLoader = (
-  {
-    graphType, data, graphTitle, labelX, labelY, width, height,
-    elementOnClick, colors, activeBiome, showOnlyTitle,
+const GraphLoader = (props) => {
+  const {
+    graphType,
+    data,
+    graphTitle,
+    colors,
+    labelX,
+    labelY,
+    width,
+    height,
+    elementOnClick,
+    activeBiome,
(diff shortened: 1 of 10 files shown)


src/charts/RenderGraph.jsx
```diff
@@ -1,23 +1,33 @@
...
- * @param {any} data Graph data, it can be null (data hasn't loaded), false (data not available)
- *  or an Object with the data.
+ * @param {any} data Graph data, it can be null (data hasn't loaded),
+ *  false (data not available) or an Object with the data.
  * @param {string} labelX axis X label
  * @param {string} labelY axis Y label
  * @param {string} graph graph type
  * @param {string} graphTitle graph title
  * @param {array} colors colors to sort elements inside the graph
  */
```
erikasv, line 17: please add documentation for the rest of the properties

cazapatamar: Done

src/charts/GraphLoader.jsx
```diff
@@ -12,27 +12,53 @@ import DotInfo from './DotInfo';
 
 ReactChartkick.addAdapter(Chart);
 
-const GraphLoader = (
-  {
-    graphType, data, graphTitle, labelX, labelY, width, height,
-    elementOnClick, colors, activeBiome, showOnlyTitle,
+const GraphLoader = (props) => {
+  const {
```
alegnaaived, line 16: Could be removed "const" and be use only {}?

alegnaaived: I tested something like this and works too:
image
image

PR #231, opened 2020-04-20
Total area box

alegnaaived:
Implement endpoint with a total area value per geofence. Styles and location should be checked with César soon.

### DIFF
--- a/src/api/RestAPI.js
+++ b/src/api/RestAPI.js
@@ -131,11 +131,11 @@ class RestAPI {
   }
 
   /**
-   * Recover area value in the selected area
+   * Recover details in the selected area
    * @param {Number} idArea id area to request, f.e. ea
    * @param {Number} idGeofence id geofence to request, f.e. idCAR
    */
-  static requestTotalAreaInGeofence(idArea, idGeofence) {
+  static requestGeofenceDetails(idArea, idGeofence) {
     return RestAPI.makeGetRequest(`${idArea}/${idGeofence}`);
   }
 
--- a/src/api/RestAPI.js
+++ b/src/api/RestAPI.js
@@ -131,6 +131,15 @@ class RestAPI {
   }
 
   /**
+   * Recover area value in the selected area
+   * @param {Number} idArea id area to request, f.e. ea
+   * @param {Number} idGeofence id geofence to request, f.e. idCAR
+   */
+  static requestTotalAreaInGeofence(idArea, idGeofence) {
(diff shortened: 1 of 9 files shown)


erikasv:
Please solve conflicts first

src/search/Drawer.jsx
```diff
@@ -60,6 +61,27 @@ class Drawer extends React.Component {
...
+            generalArea: Number(res.total_area),
+          },
+        }));
+      })
+      .catch(() => {
+        this.setState(prevState => ({
+          ...prevState,
+          data: {
+            ...prevState.data,
+            generalArea: 0,
```
erikasv, line 79: This setState in the catch isn't necessary since the initial value for generalArea is 0

src/strategicEcosystems/Overview.jsx
```diff
@@ -66,6 +67,9 @@ const Overview = (/* TODO: Add all values required */
       />
     )}
     <div className="graphcontainer pt5">
+      <GeneralArea
+        value={generalArea}
+      />
       <h4>
         hectáreas totales
         <b>{`${numberWithCommas(generalArea)} ha`}</b>
```
erikasv, line 75: Also, this isn't the same as GeneralArea component?

src/search/Drawer.jsx
alegnaaived: Should I remove all the content in the catch, as an empty catch?

erikasv: yeah, i we have an error handler that would be different, but in this case you just don't need to do anything

src/strategicEcosystems/Overview.jsx
alegnaaived: It's true. Done.

src/api/RestAPI.js
```diff
@@ -130,6 +130,15 @@ class RestAPI {
     return RestAPI.makeGetRequest(`${idArea}/${idGeofence}/se/${idSE}/pa`);
   }
 
+  /**
+   * Recover area value in the selected area
+   * @param {Number} idArea id area to request, f.e. ea
+   * @param {Number} idGeofence id geofence to request, f.e. idCAR
+   */
+  static requestTotalAreaInGeofence(idArea, idGeofence) {
```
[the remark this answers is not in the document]
alegnaaived: Done. Used requestGeofenceDetails

src/search/Drawer.jsx
alegnaaived: Done.

PR #245, opened 2020-05-06
Migrate graphs for landscape tab

cazapatamar:
Migrate graphs for landscape tab


src/charts/GraphLoader.jsx
```diff
@@ -29,29 +29,11 @@ const GraphLoader = (props) => {
     handlerInfoGraph,
     graphDescription,
     openInfoGraph,
+    zScale: zScaleInput,
```
erikasv, line 32: I think renaming zScale isn't necessary, since there is no other variable in the component with this name.

src/charts/LargeBarStackGraphNIVO.jsx
```diff
@@ -0,0 +1,119 @@
...
+    width,
+    height,
+    zScale,
+    padding,
+  } = props;
+
+  /**
+   * Give format to a big number
+   *
+   * @param {number} x number to be formatted
```
erikasv, line 19: Please add the @returns description and indicate what's the format given.

src/charts/LargeBarStackGraphNIVO.jsx
```diff
@@ -0,0 +1,119 @@
...
+   */
+  const numberWithCommas = x => x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
+
+  /**
+   * Transform data structure to be passed to component as a prop
+   *
+   * @param {array} rawData raw data from RestAPI
+   * @param {String} key id for data of each bar o serie
```
erikasv, line 27: Please add a @returns description indicating the format of the object returned

src/charts/LargeBarStackGraphNIVO.jsx
```diff
@@ -0,0 +1,119 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ResponsiveBar } from '@nivo/bar';
+
+const LargeBarStackGraphNIVO = (props) => {
```
erikasv, line 5: Is this the permanent name? I think that "NIVO" isn't necessary, I'm ok if it's only during the migration.

src/search/CompensationFactor.jsx
```diff
@@ -1,101 +1,136 @@
...
+          {openInfoGraph && (openInfoGraph === 'Factor de Compensación') && (
+            <ShortInfo
+              name="Factor de Compensación"
+              description="representa las hectáreas sobre los Biomas IAvH analizados"
+              className="graphinfo2"
+              tooltip="¿Qué significa?"
+              customButton
+            />
           )}
+          {(areaName && areaName === 'Jurisdicciones ambientales')
```
erikasv, line 53: Why don't you put all 3 graphs ("Factor de Compensación", "Biomas", "Regiones Bióticas") inside the same "if", all 3 have the same condition to be shown:

`(areaName && areaName === 'Jurisdicciones ambientales')`

erikasv: Also, since areaName is required, shouldn't be necessary to check if it's not null, only for the value:

`(areaName === 'Jurisdicciones ambientales')`

src/search/CompensationFactor.jsx
```diff
@@ -1,101 +1,136 @@
...
+                labelY="Regiones Bióticas"
+                handlerInfoGraph={handlerInfoGraph}
+                openInfoGraph={openInfoGraph}
+                graphDescription="muestra las hectáreas por cada región biótica en el área de consulta seleccionada"
+                zScale={matchColor('bioticReg')}
+                padding={0.3}
+              />
+            )
+          }
+          {(areaName && areaName !== 'Jurisdicciones ambientales')
```
erikasv, line 101: Same here, since its required it's not necessary to check it's defined

src/charts/GraphLoader.jsx
erikasv: This change is not required, but please consider it

PR #262, opened 2020-05-18
Add bar graph to current human footprint

cazapatamar:
Add bar graph to current human footprint

### DIFF
--- a/src/commons/matchColor.js
+++ b/src/commons/matchColor.js
@@ -81,11 +81,6 @@ const matchColor = (type) => {
         return palette[counter];
       };
     case 'currentHFP':
-      return (value) => {
-        const idx = sort.indexOf(value);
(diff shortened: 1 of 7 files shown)


src/commons/matchColor.js
```diff
@@ -76,6 +80,12 @@ const matchColor = (type) => {
         cache[type].counter = counter === palette.length - 1 ? 0 : counter + 1;
         return palette[counter];
       };
+    case 'currentHFP':
+      return (value) => {
+        const idx = sort.indexOf(value);
+        if (idx === -1) return palette[palette.length - 1];
+        return palette[idx];
+      };
```
erikasv, line 88: This logic is the same as coverage and biomeComp, just add the case to that one

cazapatamar: Done

PR #264, opened 2020-05-19
Move area info to header

erikasv:


erikasv:
The area information was moved to the header in both components: Search and Compensation. As required by César, logged user information was moved to a separate modal just like the login form.

Also the following changes were made:
* Fix error when there was only 1 element in the compensation selector (it was showing an very narrow autocomplete)
* Fix warnings for elements with the same key in Compensation, since there are two projects with the same name, id, and label
* Create and use context in the app for user, setUser and setHeaderNames, there will be a discussion on what to put in there and keep sending as props

PR #279, opened 2020-05-29
Feature/mock hf data

erikasv:


src/api/RestAPI.js
```diff
@@ -210,19 +204,48 @@ class RestAPI {
   /** MAPS - SEARCH MODULE */
   /** ******************** */
 
+  /**
+   * Request the geometry of the biomes by EA
+   * @param {String} eaId id ea to request
+   */
```
cazapatamar, line 210: Please add return to function description

src/api/RestAPI.js
```diff
@@ -210,19 +204,48 @@ class RestAPI {
...
+  static requestBiomesbyEAGeometry(eaId) {
+    return RestAPI.makeGetRequest(`biomes/ea/${eaId}`);
+  }
+
   /**
    * Request area geometry by id
    *
    * @param {String} areaId area id to request
    */
```
cazapatamar, line 219: Please add return to function description

src/api/RestAPI.js
```diff
@@ -210,19 +204,48 @@ class RestAPI {
...
     return {
       request: RestAPI.makeGetRequest(`${areaId}/layers/national`, { cancelToken: source.token }),
       source,
     };
   }
 
+  /**
+   * Get the geometry associated for the current footprint in the given area.
+   *
+   * @param {String} areaType area type id, f.e. "ea", "states"
```
cazapatamar, line 231: Please add return to function description and check params as apparently the function hasn't

src/api/RestAPI.js
```diff
@@ -210,19 +204,48 @@ class RestAPI {
...
+   * @param {Number | String} areaId area id to request, f.e. "CRQ", 24
+   */
+  static requestCurrentHFGeometry() {
+    return Promise.resolve(tmpCurrentHF);
+  }
+
+  /**
+   * Get the geometry associated for the footprint persistence in the given area.
+   *
+   * @param {String} areaType area type id, f.e. "ea", "states"
```
cazapatamar, line 241: Please add return to function description and check params as apparently the function hasn't

src/api/RestAPI.js
erikasv: Those are the params for the future implementation (the will be needed), I couldn't let them in the function because eslint would fail the check as they wouldn't be used right now. If you still want me to remove the doc let me know.

src/api/RestAPI.js
erikasv: Those are the params for the future implementation (the will be needed), I couldn't let them in the function because eslint would fail the check as they wouldn't be used right now. If you still want me to remove the doc let me know.

src/api/RestAPI.js
alegnaaived: I agree it should be at least a guide on how to use the returned value, isn't? Something like this? https://geojson.org/ or https://gist.github.com/cspanring/825336?short_path=365e696

src/api/RestAPI.js
cazapatamar: Now it is clear for me why the parameters descripcion is there and i agree them to stay.

src/api/RestAPI.js
erikasv: Since Angela requested changes the merge is blocked, but I believe that specification is out os scope for this case, the backend api documentation already has an explanation of what it returns and it says is a GeoJSON object, which is pretty standard.

erikasv: @cazapatamar could you please give your opinion about the example Angela is requesting?

cazapatamar: I agree is enough with the description given by the backend api documentation about the returned value.

PR #280, opened 2020-05-29
Feature/interaction multilines

alegnaaived:
It has responsive and legends according to BioTablero styles


src/commons/colorPalettes.js
```diff
@@ -92,6 +92,12 @@ export default {
     '#e66c29',
     '#cf324e',
   ],
+  currentHFPLower: [
```
erikasv, line 95: What is this palette for? It isn't being used

alegnaaived: Fixed, unused palette removed

src/charts/MultiLinesGraph.jsx
```diff
@@ -12,11 +42,13 @@ const MultiLinesGraph = (props) => {
     setSelection,
     yMin,
     yMax,
+    height,
+    width,
```
cazapatamar, line 46: Delete width prop as it is not necessary

PR #289, opened 2020-06-08
Feature/styles hfp tab

cazapatamar:


src/humanFootprint/CurrentFootprint.jsx
```diff
@@ -1,34 +1,74 @@
...
-        graphType="SmallBarStackGraph"
-        data={data}
-        units="ha"
-        colors={matchColor('currentHFP')}
-      />
+const CurrentFootprint = (props) => {
+  const {
+    data,
+    handlerInfoGraph,
+    openInfoGraph,
```
erikasv, line 12: These two properties come from Search, but nothing is being done with them. I think they don't need to be a prop, they could be an state for each component Section.

erikasv: This change would avoid re renders

src/humanFootprint/TimelineFootprint.jsx
```diff
@@ -97,17 +124,25 @@ const TimelineFootprint = (props) => {
             width="100%"
           />
         </h2>
-        Área del ecosistema dentro de la unidad de consulta: 332 ha
+        <p>
+          Área del ecosistema dentro de la unidad de consulta: 332 ha
```
alegnaaived, line 128: To review or to add an issue to call this value or hide this line in case the total area is selected

src/humanFootprint/TimelineFootprint.jsx
```diff
@@ -68,23 +69,49 @@ const changeValues = [
...
+          description="Se mostrará una gráfica de 4 líneas, una con el valor promedio de la huella en cada año para la unidad de consulta, la cual siempre estará resaltada en el gráfico, y las demás mostrarán el valor promedio en cada ecosistema estratégico"
+          className="graphinfo2"
+          tooltip="¿Qué significa?"
+          customButton
+        />
+        )
+      )}
       <h6>
-        Diferenciados entre área total y cada ecosistema estratégico
+        Huella humana comparada con EE
```
alegnaaived, line 109: Add (EE) to the accordion title

src/humanFootprint/TimelineFootprint.jsx
cazapatamar: I think we can leave like it is and implement that behavior through a new task

src/humanFootprint/TimelineFootprint.jsx
cazapatamar: Done

src/humanFootprint/CurrentFootprint.jsx
cazapatamar: The 2 props were delete from search to the rest of components. Now it is defined a state in every component to handle the behavoir to show and hide the descripcion

src/search/HumanFootprint.jsx
```diff
@@ -32,7 +32,7 @@ const HumanFootprint = ({
   },
   {
     label: {
-      id: 'Histórico y Ecosistémas estratégicos',
+      id: 'Histórico y Ecosistémas estratégicos (EE)',
```
erikasv, line 35: Please remove generalArea prop from child components render, since it was removed from them

src/humanFootprint/CurrentFootprint.jsx
erikasv: Wow, thanks for the whole refactor

src/strategicEcosystems/Overview.jsx
```diff
@@ -42,112 +42,125 @@ const helperAreaArrayIsEmpty = (array) => {
...
+            role="button"
+            tabIndex="0"
+          >
+            Área
+          </div>
+        </h2>
+        {showInfoGraph && (
+          <ShortInfo
+            name="Área"
+            description={
```
erikasv, line 95: Please let all this description in one line (airbnb styles)

src/charts/GraphLoader.jsx
```diff
@@ -136,7 +137,7 @@ const GraphLoader = (props) => {
           colors={colors}
           data={data}
           markers={markers}
-          height="490px"
+          height={490}
```
erikasv, line 140: MultilnesGraph expects an string here, either return it to string or change the propType in the graph

src/search/HumanFootprint.jsx
cazapatamar: Done

src/strategicEcosystems/Overview.jsx
cazapatamar: Done

src/charts/GraphLoader.jsx
cazapatamar: Done. Height prop has been set as number in order to standardize with other graph components

PR #301, opened 2020-06-19
Feature/current hf and persistence intereaction

cazapatamar:


erikasv:
Why, if all Human Footprint components only use the `clickOnGraph` function (`handlersGeometry[2]`) all the array is passed to them? I think it would be better just to require a function property instead of an array where only the third element is used

cazapatamar:
> left the comment in the conversation

Done

PR #302, opened 2020-06-20
Feature/tooltip on the map

cazapatamar:


src/Search.jsx
```diff
@@ -16,6 +16,20 @@ import RestAPI from './api/RestAPI';
 import matchColor from './commons/matchColor';
 import AppContext from './AppContext';
 
+/**
+ * Get the label tooltip on the map
+ */
+const getLabel = {
```
erikasv, line 22: I suggest to change this variable name, since it's not a function it shouldn't be a verb, it's kind of confusing. Could be something like "tooltipLabel" for example.

Also suggest to add a TODO to find a way to centralize this labels, right now seems a bad practice to hardcode them in two places (they are also defined in the Human Footprint components)

PR #303, opened 2020-06-26
Feature/geometry data timeline

alegnaaived:
The file src/api/tmp_hf_timeline.json was forced to have key and id. To implement NIVO component "Line" is required "id", and Search component uses "key".


src/charts/MultiLinesGraph.jsx
```diff
@@ -59,16 +59,16 @@ class MultiLinesGraph extends React.Component {
...
     });
     this.setState({ data: transformedData, selectedId });
   };
 
   selectLine = (point) => {
     const { onClickGraphHandler } = this.props;
-    this.changeSelected(point.serieId || point.id);
-    onClickGraphHandler(point.serieId || point.id);
+    this.changeSelected(point.serieId || point.id || point.key);
+    onClickGraphHandler(point.serieId || point.id || point.key);
```
erikasv, line 71: We can't just keep increasing this list size. See general comment and when we decide what to do, make the necessary change

alegnaaived: The issue is that: key has been used as the preferred standard but id is used in NIVO Line component as default. That is why I used both https://nivo.rocks/line/

Maybe we could keep key as the unique and standardized field in the response from the API and create a function before sent the data to Line component.

erikasv: That's one of the options I present in the general comment

src/api/tmp_hf_timeline.json
```diff
@@ -1,6 +1,7 @@
 [
   {
     "id": "paramo",
+    "key": "paramo",
```
erikasv, line 4: According to what we discussed, remove the id attribute

src/charts/MultiLinesGraph.jsx
alegnaaived: Opt1 Implemented

src/api/tmp_hf_timeline.json
alegnaaived: Done

src/charts/MultiLinesGraph.jsx
```diff
@@ -14,10 +14,20 @@ class MultiLinesGraph extends React.Component {
...
+    this.setData(data, colors);
+  }
+
+  /**
+   * Add fields required in data by other components, including NIVO Line (field id)
+   *
+   * @param {Object} data API response to be rendered
+   * @param {Func} colors handle color to be asigned by data type
+   */
+  setData = (data, colors) => {
```
erikasv, line 26: I don't see the necessity of this function, componentDidMount is doing nothing else, this adds complex to the component needlessly

src/charts/MultiLinesGraph.jsx
```diff
@@ -59,8 +69,8 @@ class MultiLinesGraph extends React.Component {
   changeSelected = (selectedId) => {
```
erikasv, line 69: Please add documentation to this function

PR #310, opened 2020-07-10
Feature/divide hf layer

alegnaaived:


src/Search.jsx
```diff
@@ -318,6 +320,7 @@ class Search extends Component {
     const { activeLayer } = this.state;
     const { layers } = this.state;
     const selectedSubLayer = layers[activeLayer.id].layer;
+    this.switchLayer(idCategory);
```
erikasv, line 323: This should only be called inside the ecosystems case. It's breaking the others graphs click

erikasv: Also update function documentation

src/Search.jsx
```diff
@@ -432,9 +435,9 @@ class Search extends Component {
             })
             .catch(() => this.reportDataError())
         );
-      case 'hfTimeline':
+      case 'paramo':
```
erikasv, line 438: 1. Update function documentation
2. You still need to load the persistence layer when clicking on the `hfTimeline` panel (from the accordion)

src/api/RestAPI.js
```diff
@@ -287,11 +289,21 @@ class RestAPI {
    *
    * @param {String} areaType area type id, f.e. "ea", "states"
    * @param {Number | String} areaId area id to request, f.e. "CRQ", 24
```
erikasv, line 291: You should add this two parameters to the function signature, as long as the last parameter is used, eslint won't report any error

src/api/RestAPI.js
```diff
@@ -287,11 +289,21 @@ class RestAPI {
...
+   * @param {String} seType strategic ecosystem type to request geometry
    *
    * @return {Promise<Object>} layer object to be loaded in the map
    */
-  static requestHFTimelineGeometry() {
-    return Promise.resolve(tmpHFTimelineGeo);
+  static requestHFTimelineGeometryBySE(seType) {
```
erikasv, line 296: A more appropriate name for the function would be `requestSEGeometryInGeofence` or something similar. This geometry is not attached to human footprint data

erikasv:
I noticed another thing today @alegnaaived @cazapatamar:
In the area selector, when you click in the geofence type (for example "Departamentos") the console shows the following error: "Uncaught (in promise) 500" (without line number).
I don't know where was this error introduced

src/Search.jsx
alegnaaived: Done

src/Search.jsx
alegnaaived: Done

src/api/RestAPI.js
alegnaaived: I left commented on these parameters because the order affects the function call. Hopefully, done

src/api/RestAPI.js
alegnaaived: Done

alegnaaived:
> I noticed another thing today @alegnaaived @cazapatamar:
> In the area selector, when you click in the geofence type (for example "Departamentos") the console shows the following error: "Uncaught (in promise) 500" (without line number).
> I don't know where was this error introduced

I think this should be checked for the next sprint

alegnaaived:
> 1. Don't add specific palettes and case color for each strategic ecosystem, use the correct key (`hfTimeline`) and value (`paramo`, `paramoSel`, `dryForest`, `dryForestSel`, `wetland`, `wetlandSel`) when using matchColor. If it's for map feature styling add cases where necessary.
> 2. Please remove the function "clickFeature" in Search component, which is not being used
> 3. There are multiple errors when loading search area:
>    ```
>      Warning: Failed prop type: Invalid prop `distritos` of type `boolean` supplied to `Landscape`, expected `array`.
>        in Landscape (at Drawer.jsx:306)
>        in Drawer (created by WithStyles(Drawer))
>        in WithStyles(Drawer) (at Search.jsx:798)
>        ....
>    ```
>    
>    
>    ```
>      Warning: Failed prop type: Invalid prop `bioticRegionsData` of type `boolean` supplied to `CompensationFactor`, expected `array`.
>        in CompensationFactor (at Landscape.jsx:79)
>        in Landscape (at Drawer.jsx:306)
>        ...
>    ```

2. Done

erikasv:
> > I noticed another thing today @alegnaaived @cazapatamar:
> > In the area selector, when you click in the geofence type (for example "Departamentos") the console shows the following error: "Uncaught (in promise) 500" (without line number).
> > I don't know where was this error introduced
> 
> I think this should be checked for the next sprint

Then create a task for it

src/api/RestAPI.js
```diff
@@ -255,21 +257,38 @@ class RestAPI {
...
-   * Request area geometry by id
+   * Request area geometry by id divided by geofences
    *
    * @param {String} areaId area id to request
    *
    * @return {Object} Including Promise with layer object to load in map and source reference to
    * cancel the request
    */
-  static requestGeofenceGeometry(areaId) {
+  static requestNationalGeometryByArea(areaId) {
```
erikasv, line 267: You didn't change the function name where it's called

erikasv:
> > 1. Don't add specific palettes and case color for each strategic ecosystem, use the correct key (`hfTimeline`) and value (`paramo`, `paramoSel`, `dryForest`, `dryForestSel`, `wetland`, `wetlandSel`) when using matchColor. If it's for map feature styling add cases where necessary.
> > 2. Please remove the function "clickFeature" in Search component, which is not being used
> > 3. There are multiple errors when loading search area:
> >    ```
> >      Warning: Failed prop type: Invalid prop `distritos` of type `boolean` supplied to `Landscape`, expected `array`.
> >        in Landscape (at Drawer.jsx:306)
> >        in Drawer (created by WithStyles(Drawer))
> >        in WithStyles(Drawer) (at Search.jsx:798)
> >        ....
> >    ```
> >    
> >    
> >    ```
> >      Warning: Failed prop type: Invalid prop `bioticRegionsData` of type `boolean` supplied to `CompensationFactor`, expected `array`.
> >        in CompensationFactor (at Landscape.jsx:79)
> >        in Landscape (at Drawer.jsx:306)
> >        ...
> >    ```
> 
>     1. I need help to do this
> 
>     2. Done
> 
>     3. I think this should be checked for next sprint

I didn't see this on friday.

1. I can do a commit replacing those calls, but first fix the other problems so I can understand why did you added those cases.

3. No, that problem must be caused on the merge of your branch with develop, because when Camilo created Landscape component those error weren't happening

erikasv:
> > > 1. Don't add specific palettes and case color for each strategic ecosystem, use the correct key (`hfTimeline`) and value (`paramo`, `paramoSel`, `dryForest`, `dryForestSel`, `wetland`, `wetlandSel`) when using matchColor. If it's for map feature styling add cases where necessary.
> > > 2. Please remove the function "clickFeature" in Search component, which is not being used
> > > 3. There are multiple errors when loading search area:
> > >    ```
> > >      Warning: Failed prop type: Invalid prop `distritos` of type `boolean` supplied to `Landscape`, expected `array`.
> > >        in Landscape (at Drawer.jsx:306)
> > >        in Drawer (created by WithStyles(Drawer))
> > >        in WithStyles(Drawer) (at Search.jsx:798)
> > >        ....
> > >    ```
> > >    
> > >    
> > >    ```
> > >      Warning: Failed prop type: Invalid prop `bioticRegionsData` of type `boolean` supplied to `CompensationFactor`, expected `array`.
> > >        in CompensationFactor (at Landscape.jsx:79)
> > >        in Landscape (at Drawer.jsx:306)
> > >        ...
> > >    ```
> > 
> > 
> > ```
> > 1. I need help to do this
> > 
> > 2. Done
> > 
> > 3. I think this should be checked for next sprint
> > ```
> 
> I didn't see this on friday.
> 
>     1. I can do a commit replacing those calls, but first fix the other problems so I can understand why did you added those cases.
> 
>     2. Ok
> 
>     3. No, that problem must be caused on the merge of your branch with develop, because when Camilo created Landscape component those error weren't happening

Sorry about the 3rd point, you are right, the error it's older than the merge, so, please add it to the task for the other error.

src/Search.jsx
```diff
@@ -319,23 +310,22 @@ class Search extends Component {
...
     selectedSubLayer.eachLayer((layer) => {
+      // 'switch' for strategic ecosystem selected
+      switch (idCategory) {
+        case 'paramo':
+        case 'wetland':
+        case 'dryForest':
+          this.switchLayer(idCategory);
+          this.setTimelineHFData(tooltipLabel[idCategory]);
+          break;
+        // no default
```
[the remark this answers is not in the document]
alegnaaived: Well, I just did this according to the linter suggestion

src/api/RestAPI.js
```diff
@@ -255,21 +257,38 @@ class RestAPI {
   }
 
   /**
-   * Request area geometry by id
+   * Request area geometry by id divided by geofences
```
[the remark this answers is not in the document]
alegnaaived: Done.

src/api/RestAPI.js
```diff
@@ -287,11 +306,21 @@ class RestAPI {
    *
    * @param {String} areaType area type id, f.e. "ea", "states"
    * @param {Number | String} areaId area id to request, f.e. "CRQ", 24
+   * @param {String} seType strategic ecosystem type to request geometry
    *
    * @return {Promise<Object>} layer object to be loaded in the map
    */
-  static requestHFTimelineGeometry() {
-    return Promise.resolve(tmpHFTimelineGeo);
+  static requestSEGeometryInGeofence(/* areaType, areaId, */ seType) {
```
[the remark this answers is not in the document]
alegnaaived: Done.

src/api/RestAPI.js
alegnaaived: Thank you, I fixed this. Done.

src/Search.jsx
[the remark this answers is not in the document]
erikasv: Well, the rule actually says that you should follow a code convention: https://eslint.org/docs/rules/default-case
And that comment is an accepted convention, as well as the empty default case, we as a team should agree in following only one.
I don't like the comment convention because you already use a lot of comments (like the ones before the switch and before the if of this function) so it's not clear that this particular comment is a convention we should follow.
@cazapatamar we need your opinion, if both of you prefer the comment convention it's ok for me, I just want that all switch cases that don't have a default behavior use the same convention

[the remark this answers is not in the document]
cazapatamar: I like better avoid using all kind of comments in the code. Therefore, i agree to always have a default case even if it is empty or return null. @erikasv @alegnaaived

src/api/RestAPI.js
```diff
@@ -283,15 +302,26 @@ class RestAPI {
...
-    return Promise.resolve(tmpHFTimelineGeo);
+  static requestHFGeometryBySEInGeofence(/* areaType, areaId, */ seType) {
+    switch (seType) {
+      case 'dryForest':
+        return Promise.resolve(tmpHFTDryForestGeo);
+      case 'paramo':
+        return Promise.resolve(tmpHFTParamoGeo);
+      case 'wetland':
+        return Promise.resolve(tmpHFTWetlandGeo);
+      // no default
```
erikasv, line 322: please add a default empty case

src/Search.jsx
```diff
@@ -319,23 +310,22 @@ class Search extends Component {
     const { layers } = this.state;
     const selectedSubLayer = layers[activeLayer.id].layer;
     selectedSubLayer.eachLayer((layer) => {
```
erikasv, line 312: I don't think this is the correct way to handle the changes for this graph. setTimelineHFData is being called a lot of times, like 7, this causes the request for every ecosystem to be made that amount of times and that's a problem.

I'm helping you to find out how can be handled.

erikasv:
The problem extends to the layers state, we are using only the geofence id as the key for the layers, so we are replacing all of them with each request, but we should fix it in other issue.

To avoid conflicts, for now just get the switch out of the `eachLayer`, this works when "illuminating" a feature, but as the switch works for loading new information as a whole layer, it's doing the same the amount of features in the current selected layer.

This is the state of the function which reduced the request to one for me:
```js
clickOnGraph = (idCategory) => {
    const { activeLayer } = this.state;
    const { layers } = this.state;
    const selectedSubLayer = layers[activeLayer.id].layer;
    selectedSubLayer.eachLayer((layer) => {
      if (layer.feature.properties.key === idCategory) {
        layer.setStyle({
          weight: 1,
          fillOpacity: 1,
        });
      } else {
        selectedSubLayer.resetStyle(layer);
      }
    });
    switch (idCategory) {
      case 'paramo':
      case 'wetland':
      case 'dryForest':
        this.switchLayer(idCategory);
        this.setTimelineHFData(tooltipLabel[idCategory]);
        break;
      default:
        break;
    }
  };
```

src/Search.jsx
erikasv: See conversation comment with a solution (i think it could be handled better, but for now it's ok)

src/api/RestAPI.js
alegnaaived: Done.

PR #313, opened 2020-07-14
Change HFP by HF references on the code

cazapatamar:


src/landscape/Landscape.jsx
```diff
@@ -123,7 +123,7 @@ Landscape.propTypes = {
   biomas: PropTypes.array,
   distritos: PropTypes.array,
   currentHF: PropTypes.array,
-  currentHFPValue: PropTypes.number,
+  currentHFValue: PropTypes.number,
```
alegnaaived, line 126: Could you please change it to hfCurrentValue?

src/landscape/Landscape.jsx
```diff
@@ -96,7 +96,7 @@ class Landscape extends React.Component {
         component: (
           <HumanFootprint
             currentHF={currentHF}
-            currentHFPValue={currentHFPValue}
+            currentHFValue={currentHFValue}
             hfPersistence={hfPersistence}
             hfTimeline={hfTimeline}
             timelineHFArea={timelineHFArea}
```
alegnaaived, line 102: It makes me think to change it too, to keep hfTimelineArea to standardize the beginning as "hf"

src/Search.jsx
```diff
@@ -98,7 +98,7 @@ class Search extends Component {
   };
 
   /**
-   * Set in state timelineHFP area details for strategic ecosystems (SE) in the selected area
+   * Set in state timelineHF area details for strategic ecosystems (SE) in the selected area
```
alegnaaived, line 101: Then, it should be better hfTimelineArea instead of timelineHF area. Thank you for change this

cazapatamar: Done

PR #327, opened 2020-07-29
Map adjustments and connect strategic ecosystems layers with endpoints

erikasv:


src/Search.jsx
```diff
@@ -600,25 +546,29 @@ class Search extends Component {
         },
       );
     }
+    if (nameToOff) {
+      this.setState((prevState) => {
+        const newState = { ...prevState };
+        newState.layers[nameToOff].active = false;
```
alegnaaived, line 552: I got this error from "Consultas" to "Areas de manejo especial":

image

PR #336, opened 2020-08-24
Feature: Map title

erikasv:
Add map title. Additionally some other adjustments were made:

* Refactor switchLayer function to reduce duplicated code
* Load layers from state instead o requesting the again

### DIFF
--- a/src/Search.jsx
+++ b/src/Search.jsx
@@ -344,6 +344,7 @@ class Search extends Component {
       requestSource,
       selectedArea,
       selectedAreaType,
+      layers,
     } = this.state;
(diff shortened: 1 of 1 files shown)


cazapatamar:
If the "Huella Humana" tab is open and the 3 sub-tabs are closed (Current, persistence and timeline) the title shows the last sub-tab opened. I think that in that case should not be a title on the map.

cazapatamar:
Also if i back to "ecosistemas" or "especies" tab, in the same case the title on the map shows the title for the last opened layer. Should be empty.

PR #338, opened 2020-08-27
Feature/default map

alegnaaived:


src/api/RestAPI.js
```diff
@@ -295,15 +295,10 @@ class RestAPI {
...
    *
-   * @return {Object} Including Promise with layer object to load in map and source reference to
-   * cancel the request
+   * @return {Promise<Object>} layer object to be loaded in the map
    */
   static requestGeofenceGeometryByArea(areaId, geofenceId) {
-    const source = CancelToken.source();
-    return {
-      request: RestAPI.makeGetRequest(`${areaId}/layers/${geofenceId}`, { cancelToken: source.token }),
-      source,
```
erikasv, line 305: Why did you remove this? I actually think every geometry request should return the cancel token. Once the loading modal is attached only to the map, we'll need to cancel requests

alegnaaived: Okay, I will keep this to cancel the request

src/api/RestAPI.js
```diff
@@ -342,15 +353,28 @@ class RestAPI {
...
       case 'wetland':
-        return RestAPI.makeGetRequest(`${areaType}/${areaId}/se/layers/Humedal`);
+        return {
+          request: RestAPI.makeGetRequest(`${areaType}/${areaId}/se/layers/Humedal`, { cancelToken: source.token }),
+          source,
+        };
       default:
-        return undefined;
+        return {
+          request: undefined,
```
erikasv, line 375: Please change this for `Promise.reject(new Error('undefined option))` to avoid problems in switchLayer function

erikasv (changes requested):

src/api/RestAPI.js
alegnaaived: Done

src/api/RestAPI.js
alegnaaived: Done (with your help) (y)

erikasv (approved):

PR #341, opened 2020-09-10
Change loading modal to capture only MapViewer

erikasv:


cazapatamar (commented):
Great behaviour. Just a question, as the modal for dataError show up when there is not features in received response, i mean, it is also oriented to geometry data, should also we move this modal to MapViewer? What do you think?

erikasv:
> Great behaviour. Just a question, as the modal for dataError show up when there is not features in received response, i mean, it is also oriented to geometry data, should also we move this modal to MapViewer? What do you think?

Great point, maybe we should. But also we should change the content to "Capa no disponible actualmente" or something like that? since the current message would be too general.

cazapatamar:
> > Great behaviour. Just a question, as the modal for dataError show up when there is not features in received response, i mean, it is also oriented to geometry data, should also we move this modal to MapViewer? What do you think?
> 
> Great point, maybe we should. But also we should change the content to "Capa no disponible actualmente" or something like that? since the current message would be too general.

Agree with the message "Capa no disponible actualmente"

cazapatamar (approved):

PR #356, opened 2020-10-29
State loadingLayer = false when backbutton

cazapatamar:


erikasv:
Please change the branch name to feature/... so the actions can run.

In some future (hopefully soon) we'll add checks for other type of branches

cazapatamar:
> Please change the branch name to feature/... so the actions can run.
> 
> In some future (hopefully soon) we'll add checks for other type of branches

If i change the name locally and push it again, GitHub will create a new branch, right? Then, shuould i create a new PR or there is a way to link the new branch to this PR?

erikasv:
Noup, close this one and create a new one

cazapatamar:
Closed because the branch name is wrong according to CI definitions

PR #371, opened 2020-11-11
Loss and persistence forest graph inside accordion

cazapatamar:


erikasv (changes requested):
ForestLossPersistence component is missing the following:

* componentWillUnmount function (`mounted` variable never is false)
* mounted variable is not declared in the class

erikasv (approved):

PR #372, opened 2020-11-12
Landscape texts

cazapatamar:


cazapatamar:
Ready for review, finally i left the condition for specific color and later we could implement a better option. @erikasv @Trjegul84

erikasv (approved):

PR #382, opened 2020-11-27
Forest loss and persistence geometries

cazapatamar:


src/Search.jsx
```diff
@@ -474,6 +511,17 @@ class Search extends Component {
...
+        layerKey = 'forestLP-2016-2019';
+        request = () => RestAPI.requestEcoChangeLPGeometry(
+          selectedAreaTypeId, selectedAreaId, '2016-2019',
+        );
+        layerStyle = this.featureStyle({ type: 'forestLP' });
+        newActiveLayer = {
+          id: 'forestLP-2016-2019',
+          name: 'Pérdida y persistencia de bosque (2016-2019)',
+        };
+        break;
```
erikasv, line 524: I think this case isn't being used?

erikasv (changes requested):
Sorry, I didn't realize it was marked as ready for review.

src/Search.jsx
cazapatamar: Fixed

erikasv (approved):

PR #401, opened 2021-02-22
deactivate ForestLossPersistence accordion and set Integrity section …

Trjegul84:
New property called deactivated was set for deactivate accordion in Lanscape componentsArray.


cazapatamar (approved):

cazapatamar:
Don't forget to connect this PR with the corresponding issue

erikasv (changes requested):
4 things:
1. Why not to reuse the `disabled` attribute?
1. In any case, it can be false by default, so it's not necessary to always pass it, besides it should be the default behavior
1. Fix PropType in  LandscapeAccordion, right now it's documented as `disable`
1. In case is necessary to keep both, document it on PropTypes

Trjegul84:
I think there is a kind of misnomer in props names because the **disabled** defined in componentsArray is related to collapsed or expanded accordion and the new **deactivated** is for setting the disabled prop for Material UI Accordion furthermore in Landscape.jsx there is another state named **visible** that seems to be used to identify the visible (expanded) accordion by default.  Should we re-assign the names?

Trjegul84:
The deactivated property was deleted. New property collapsed will replace old disabled, now we use disabled to mean inactive (disabled/not visible). Collapsed and new disabled will be optional and both set in false by default. If collapsed is true, the tab will not be expanded and the first tab displayed to user will be first without collapsed property.  In order to disabled any tab (set disabled=true) it must have collapsed in true (not expanded) otherwise it will not function properly.

erikasv (approved):

PR #410, opened 2021-03-02
scihf endpoints

erikasv:


cazapatamar (approved):

Trjegul84 (commented):
Don't think if these are related with current changes but i found the following:

. When select department, then in forest tab/ integrity, if the layer is fully loaded and you clic in pie the layer everything is ok but if map is not fully loaded if you clic in pie the following error is displayed. 

Screenshot_20210304_172026

It happens with CAR also if you select other sector in pie.

- When special areas are selected the forest integrity pie is displayed but map doesn't other times the message no available layers is displayed but it shows weird empty pie.

Screenshot_20210304_173847

Screenshot_20210304_173951

erikasv:
> Don't think if these are related with current changes but i found the following:
> 
> . When select department, then in forest tab/ integrity, if the layer is fully loaded and you clic in pie the layer everything is ok but if map is not fully loaded if you clic in pie the following error is displayed.

Yeah, I already created an [issue](https://github.com/PEM-Humboldt/biotablero-frontend/issues/409) for that.

erikasv:
>     * When special areas are selected the forest integrity pie is displayed but map doesn't other times the message no available layers is displayed but it shows weird empty pie.

The map may be displayed but the layer too disperse. Some special areas have problems with the data, if you have the exact special areas we can check with Jaime if everything is ok with the data.

Also in case there is simply no data we should control that on. @cazapatamar do you remember what we did for the human footprint timeline in that case?

Trjegul84:
@erikasv  you are right probably there are tiny layers it can be seen. Areas de Recreación and Santuario de Fauna y Flora are two examples.

Trjegul84 (approved):

cazapatamar:
> > ```
> > * When special areas are selected the forest integrity pie is displayed but map doesn't other times the message no available layers is displayed but it shows weird empty pie.
> > ```
> 
> The map may be displayed but the layer too disperse. Some special areas have problems with the data, if you have the exact special areas we can check with Jaime if everything is ok with the data.
> 
> Also in case there is simply no data we should control that on. @cazapatamar do you remember what we did for the human footprint timeline in that case?

I will check with 2 small areas that Angela provides. Maybe the endpoint returns a empty array instead of nothing (null) and nivo could be trying to render that array.

PR #412, opened 2021-03-08
Connectivity section1 graphs

cazapatamar:


Trjegul84 (approved):

src/pages/search/drawer/landscape/connectivity/CurrentPAConnectivity.jsx
```diff
@@ -28,12 +74,71 @@ class CurrentPAConnectivity extends React.Component {
...
+        <div>
+          <h6>
+            Conectividad áreas protegidas
+          </h6>
+          <div>
+            <h6>
+              Indice Prot
+            </h6>
+            <h5 style={{ backgroundColor: '#d5753d' }}>
+              {prot}
```
erikasv, line 111: this is a percentage right? (based on line 46) it should have the `%` symbol or a more explicit name. Or is it `Indice Prot` a percentage in general?.

I thought this value would be the total prot area :sweat_smile:

erikasv (changes requested):
The layer is not being loaded (isn't even called)

cazapatamar:
> The layer is not being loaded (isn't even called)

The layer and interaction for this section is included in other issue. I have planned work on it in other PR

PR #425, opened 2021-03-16
display geometry in connectivity section 2

Trjegul84:
Only call add timelineConn cases to apply same functions of section 1 in section 2 also same colors.

### DIFF
--- a/src/pages/Search.jsx
+++ b/src/pages/Search.jsx
@@ -541,10 +541,10 @@ class Search extends Component {
           5,
         );
         shutOtherLayers = false;
-        layerStyle = this.featureStyle({ type: 'currentPAConn' });
+        layerStyle = this.featureStyle({ type: 'currentPAConn', fKey: 'dpc_cat' });
         layerKey = 'currentPAConn';
         newActiveLayer = {
-          id: 'layerType',
+          id: 'currentPAConn',
           name: 'Histórico de conectividad áreas protegidas',
         };
         break;
--- a/src/pages/Search.jsx
+++ b/src/pages/Search.jsx
@@ -184,7 +184,7 @@ class Search extends Component {
   featureStyle = ({ type, color = null, fKey = 'key' }) => (feature) => {
     if (feature.properties) {
       const key = type === 'fc' ? feature.properties.compensation_factor : feature.properties[fKey];
-      const ftype = type === 'currentPAConn' ? 'dpc' : type;
+      const ftype = (type === 'currentPAConn' || type === 'timelinePAConn') ? 'dpc' : type;
       if (!key) {
         return {
           color: matchColor(ftype)(color),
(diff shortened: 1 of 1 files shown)


src/pages/Search.jsx
```diff
@@ -519,6 +520,7 @@ class Search extends Component {
         };
         break;
       case 'currentPAConn':
+      case 'timelinePAConn':
```
erikasv, line 523: You need to make a separated case because of the layer name

erikasv (changes requested):

erikasv:
@cazapatamar can you check the geofence layer is being shown? In my local it's showing only the dpc layer in both cases (current and timeline)

cazapatamar:
> @cazapatamar can you check the geofence layer is being shown? In my local it's showing only the dpc layer in both cases (current and timeline)

In my local i see both layers

image

Trjegul84:
Using develop branch in my local I see both layers are displayed in current and timeline. I thought this was due to mockup data is not linked with area_id. 
Screenshot_20210129_105602

src/pages/Search.jsx
```diff
@@ -532,6 +533,20 @@ class Search extends Component {
...
+        this.switchLayer('geofence');
+        request = () => RestAPI.requestDPCLayer(
+          selectedAreaTypeId,
+          selectedAreaId,
+          5,
+        );
+        shutOtherLayers = false;
+        layerStyle = this.featureStyle({ type: layerType, fKey: 'dpc_cat' });
+        newActiveLayer = {
+          id: layerType,
```
erikasv, line 546: Use `currentPAConn` as the active layer id and also as the `layerKey`, so you don't load twice the same layer.

See the case for hf timeline for reference

erikasv (changes requested):

src/pages/Search.jsx
Trjegul84: Ok. Changed. It is better solution than duplicate the code.

src/pages/Search.jsx
```diff
@@ -532,6 +533,21 @@ class Search extends Component {
...
+        request = () => RestAPI.requestDPCLayer(
+          selectedAreaTypeId,
+          selectedAreaId,
+          5,
+        );
+        shutOtherLayers = false;
+        layerStyle = this.featureStyle({ type: 'currentPAConn' });
+        layerKey = 'currentPAConn';
+        newActiveLayer = {
+          id: 'layerType',
```
erikasv, line 547: You just set the new active layer id to the string "layerType", it should be `'currentPAConn'` as it's reusing the layer for currentPAConn

src/pages/Search.jsx
```diff
@@ -532,6 +533,21 @@ class Search extends Component {
...
         break;
+      case 'timelinePAConn':
+        this.switchLayer('geofence');
+        request = () => RestAPI.requestDPCLayer(
+          selectedAreaTypeId,
+          selectedAreaId,
+          5,
+        );
+        shutOtherLayers = false;
+        layerStyle = this.featureStyle({ type: 'currentPAConn' });
```
erikasv, line 544: you must send `fKey` or it won't work, right now it works because the layer loads too fast, but if you change accordion tab before the layer for the first section is loaded, this won't load correctly

erikasv (changes requested):

erikasv (approved):

PR #428, opened 2021-03-26
Geometry pa connectivity section 3

cazapatamar:


src/pages/Search.jsx
```diff
@@ -767,6 +807,7 @@ class Search extends Component {
       'forestIntegrity',
       'currentPAConn',
       'timelinePAConn',
+      'currentSEPAConn',
```
erikasv, line 810: you're missing cleaning the strategic ecosystems layers

src/utils/colorPalettes.js
```diff
@@ -86,6 +86,9 @@ export default {
   paramo: ['#eaa04a'],
   dryForest: ['#8cb64c'],
   wetland: ['#6794cb'],
+  paramoPAConn: ['#120000'],
+  dryForestPAConn: ['#120000'],
+  wetlandPAConn: ['#120000'],
```
erikasv, line 91: these can be grouped into one palette since they are the same

src/utils/matchColor.js
```diff
@@ -65,6 +65,18 @@ const match = {
...
+    sort: ['paramoPAConn'],
+  },
+  dryForestPAConn: {
+    palette: 'dryForestPAConn',
+    sort: ['dryForestPAConn'],
+  },
+  wetlandPAConn: {
+    palette: 'wetlandPAConn',
+    sort: ['wetlandPAConn'],
+  },
```
erikasv, line 79: In all these cases you don't need to define a sort since you are using the default behavior.

erikasv (changes requested):

src/utils/matchColor.js
cazapatamar: Done

src/utils/colorPalettes.js
cazapatamar: Done

src/pages/Search.jsx
cazapatamar: Done

erikasv (approved):

Trjegul84 (approved):

PR #431, opened 2021-04-05
Connectivity visual adjustments

cazapatamar:
Section 2 map shows all PAs instead of top five and set a proper name to SE layer by clicking graph in section 3 according to selected SE

### DIFF
--- a/src/pages/Search.jsx
+++ b/src/pages/Search.jsx
@@ -573,9 +573,8 @@ class Search extends Component {
         );
         shutOtherLayers = false;
         layerStyle = this.featureStyle({ type: 'currentPAConn', fKey: 'dpc_cat' });
-        layerKey = 'timelinePAConn';
         newActiveLayer = {
-          id: 'currentPAConn',
+          id: 'timelinePAConn',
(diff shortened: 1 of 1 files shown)


Trjegul84 (approved):

src/pages/Search.jsx
```diff
@@ -570,11 +570,10 @@ class Search extends Component {
...
           selectedAreaTypeId,
           selectedAreaId,
-          5,
         );
         shutOtherLayers = false;
         layerStyle = this.featureStyle({ type: 'currentPAConn', fKey: 'dpc_cat' });
-        layerKey = 'currentPAConn';
+        layerKey = 'timelinePAConn';
         newActiveLayer = {
           id: 'currentPAConn',
```
erikasv, line 578: Since there is no interaction between chart and map this won't be an issue, but just in case, to avoid a future hard to find bug, you should change this id, since this layer no longer is the same as "currentPAConn"

cazapatamar: layerKey assignment was removed and newActiveLayer id changed to timelinePAConn

erikasv (approved):

PR #440, opened 2021-04-13
Forest integrity changes

erikasv:


Trjegul84 (approved):
The highlight colors look much better and are much more visible in the map. However the "Distribución en áreas protegidas" SmallBarStackGraph bar still has the same colors as before, should this be like the case?

erikasv:
> The highlight colors look much better and are much more visible in the map. However the "Distribución en áreas protegidas" SmallBarStackGraph bar still has the same colors as before, should this be like the case?

I understand it is. Those colors are the same as "áreas protegidas" in ecosystems tab, also César worked with me in this PR so he saw those colors again xD

cazapatamar (approved):

PR #445, opened 2021-04-20
Final styles connectivity module

cazapatamar:


Trjegul84 (commented):
This comment is separate from the styles because in the previous review I did't see the case when no information available message is displayed. 
I would think is right if there is no info in database to differ from case of zero protected areas but I don't know if this is what is agreed for this case. 

Screenshot_20210420_185819

Trjegul84 (approved):

cazapatamar:
> This comment is separate from the styles because in the previous review I did't see the case when no information available message is displayed.
> I would think is right if there is no info in database to differ from case of zero protected areas but I don't know if this is what is agreed for this case.
> 
> Screenshot_20210420_185819

As we agreed in last friday meeting, the message of unavailable information is shown both for no data and zero values

erikasv (changes requested):
Please update the spreadsheet with color palettes, complete `Gráficos o geometría a los que aplica` column and remove the ones that don't exist on `colorPalettes.js` file

cazapatamar:
> Please update the spreadsheet with color palettes, complete `Gráficos o geometría a los que aplica` column and remove the ones that don't exist on `colorPalettes.js` file

Done

PR #458, opened 2021-05-03
Feature/fix nivo bug

Trjegul84:
Bug is cleared by upgrading Nivo libraries to 0.69.0 version then react-spring dependency is related to 9.1.2.


erikasv (changes requested):
Forest integrity chart is broken, maybe prop names chaged.

erikasv:
> Forest integrity chart is broken, maybe prop names chaged.

PR #460, opened 2021-05-03
New adjustments in connectivity

cazapatamar:


Trjegul84 (changes requested):
I've highlight in red some values that are not completely visible, I think the numbers are to long to be there and not placed well. Are those necessary at the left of the bars?

Cifras-NoVisibles

cazapatamar:
> I've highlight in red some values that are not completely visible, I think the numbers are to long to be there and not placed well. Are those necessary at the left of the bars?
> 
> Cifras-NoVisibles

It seems to be a problem with nivo upgrade you did in the other PR. Try to delete node_modules, install dependencies again been on this branch and test again. In my local env i don't see labels at Y-axis as it should be.

image

if indeed it is issue related to nivo upgrade, please solve it in the PR #459

Trjegul84:
Ok, thanks, I will check again.

erikasv:
@cazapatamar are you suggesting we should merge this branch first?, because if that problem happens because of the nivo upgrade it equally has to be solved

cazapatamar:
> @cazapatamar are you suggesting we should merge this branch first?, because if that problem happens because of the nivo upgrade it equally has to be solved

I mean, first solve the issue but in the other PR where we are upgrading nivo and not here. What do you think? @erikasv

erikasv:
> I mean, first solve the issue but in the other PR where we are upgrading nivo and not here. What do you think? @erikasv

Yeah agree, I just saw your comment in the other PR. I just was worried the problem wouldn't be solved.

cazapatamar:
> > I mean, first solve the issue but in the other PR where we are upgrading nivo and not here. What do you think? @erikasv
> 
> Yeah agree, I just saw your comment in the other PR. I just was worried the problem wouldn't be solved.

@erikasv After checking with @Trjegul84 we realize the issue is not associated to nivo upgrade but to an error inside the component as i was setting the axisLeft prop instead of being null. It was already fixed in the last commit.

src/components/charts/GraphLoader.jsx
```diff
@@ -92,7 +92,6 @@ const GraphLoader = (props) => {
           onClickHandler={onClickGraphHandler}
           selectedIndexValue={selectedIndexValue}
           labelX={labelX}
-          labelY={labelY}
```
erikasv, line 95: Wouldn't this affect the Loss and persistence graph?

src/components/charts/MultiSmallSingleBarGraph.jsx
```diff
@@ -20,7 +20,6 @@ class MultiSmallSingleBarGraph extends React.Component {
       units,
       onClickHandler,
       labelX,
-      labelY,
```
erikasv, line 23: Same that previous comment, this wouldn't affect loss and persistence section?

src/components/charts/GraphLoader.jsx
cazapatamar: No, the component is different for Loss and persistence graph whose name is MultiSmallBarStackGraph

src/components/charts/MultiSmallSingleBarGraph.jsx
cazapatamar: Same last comment

Trjegul84 (approved):

erikasv (approved):

PR #465, opened 2021-05-05
Species accordions and empty components

cazapatamar:


Trjegul84 (approved):

src/pages/search/drawer/Species.jsx
```diff
@@ -0,0 +1,112 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import Richness from 'pages/search/drawer/species/Richness';
+import FunctionalDiversity from 'pages/search/drawer/species/FunctionalDiversity';
+import Accordion from 'pages/search/drawer/Accordion';
+import SearchContext from 'pages/search/SearchContext';
+
+class Landscape extends React.Component {
```
erikasv, line 9: I believe this class should be Species?

erikasv: Also I see this component too similar to Landscape... maybe in the future we should evaluate the possibility to unify them in one component

erikasv (changes requested):

src/pages/search/drawer/Species.jsx
cazapatamar: Fixed the Species component name.

On the other hand, you are right, this component is pretty similar to Landscape. It is a good idea to make it one component to handle Species and Landscape first level accordions.

erikasv (approved):

PR #466, opened 2021-05-12
Richness -> number of species

erikasv:


src/components/charts/GraphLoader.jsx
```diff
@@ -105,6 +104,15 @@ const GraphLoader = (props) => {
           onClickHandler={onClickGraphHandler}
         />
       );
+    case 'singleBullet':
+      return (
+        <SingleBulletGraph
+          data={data}
+          height={100}
+          colors={colors}
+          onClickHandler={onClickGraphHandler}
```
cazapatamar, line 113: onClickHandler is not used (for now) inside the component. Should we deleted while we define interactions?

cazapatamar:
There is something weird with responsiveness. If resize the bars width dont fit properly in the section. @Trjegul84 @alegnaaived do you have the same behaviour.

Trjegul84:
> There is something weird with responsiveness. If resize the bars width dont fit properly in the section. @Trjegul84 @alegnaaived do you have the same behaviour.
> 
Yes, I have the same behavior.

Trjegul84 (commented):
Also when you resize the window something happend in the left margin of the first bar.

Screenshot_20210512_131408

alegnaaived:
> There is something weird with responsiveness. If resize the bars width dont fit properly in the section. @Trjegul84 @alegnaaived do you have the same behaviour.
> 
> image

I couldn't get the same error but check the blue alignment in the blue legend

alegnaaived:
I think we should check eventually legend styles with Cesar because the size and font at the top and bottom could be confusing for users

erikasv:
> I think we should check eventually legend styles with Cesar because the size and font at the top and bottom could be confusing for users

Those aren't the final legends (neither their layout), legends, fonts and other style stuff are always modified later with Cesar, at the end when we have definitive color palette and real data to be sure the proposed chart and color fit ok.

src/components/charts/GraphLoader.jsx
erikasv: Yeah you're right, we don't even know if this chart will have interaction with the map

erikasv:
The responsive problem was actually caused by the custom components, because they were not responsive :sweat_smile: 

To solve this I tried to "copy" some of the default components nivo uses, but this required to use react-spring (and this dep is still **black box** for me a :sweat_smile: )

cazapatamar (approved):

Trjegul84:
According to what some people express in the forums, the only way to fix the responsive issues is to install react-spring or something like a patch of react-spring and probably this will help other Nivo graphics to be used in the future to behave well when window is resized.

Trjegul84 (approved):

PR #476, opened 2021-05-24
add  y axis limit for TimelinePAConnectivity graph

Trjegul84:
With this changes, the limit for y-axis (yMax) in timelinePAConnectivity graph is 50. The y-axis limit for HH remains as default value.

### DIFF
--- a/src/components/charts/GraphLoader.jsx
+++ b/src/components/charts/GraphLoader.jsx
@@ -29,6 +29,7 @@ const GraphLoader = (props) => {
     markers,
     loading,
     selectedIndexValue,
+    yMax,
   } = props;
 
   // While data is being retrieved from server
@@ -161,6 +162,7 @@ const GraphLoader = (props) => {
           markers={markers}
           height={490}
           units={units}
+          yMax={yMax}
         />
       );
     default:
--- a/src/components/charts/GraphLoader.jsx
+++ b/src/components/charts/GraphLoader.jsx
@@ -183,6 +183,7 @@ GraphLoader.propTypes = {
   width: PropTypes.number,
   showOnlyTitle: PropTypes.bool,
   units: PropTypes.string,
+  yMax: PropTypes.number,
   elementOnClick: PropTypes.func,
   // TODO: Remove array type once the charts in compensation are migrated
   colors: PropTypes.oneOfType([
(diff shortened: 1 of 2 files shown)


src/components/charts/GraphLoader.jsx
```diff
@@ -29,6 +29,7 @@ const GraphLoader = (props) => {
     markers,
     loading,
     selectedIndexValue,
+    yMax,
```
erikasv, line 32: Document the prop

erikasv (changes requested):
What is this change for?

cazapatamar:
This is for ea=CAR

image

src/pages/search/drawer/landscape/connectivity/TimelinePAConnectivity.jsx
```diff
@@ -17,6 +17,7 @@ class TimelinePAConnectivity extends React.Component {
     this.state = {
       showInfoGraph: false,
       timelinePAConnectivity: [],
+      yLimit: 50,
```
cazapatamar, line 20: Why use this value as a state instead of only send the value in line 120?

Trjegul84:
@cazapatamar I don't see the graph like you. This is how it looks like for me:
Screenshot_20210524_151319

cazapatamar (changes requested):

cazapatamar:
> @cazapatamar I don't see the graph like you. This is how it looks like for me:
> Screenshot_20210524_151319

Sorry, my mistake for being at the wrong backend branch.

src/components/charts/GraphLoader.jsx
Trjegul84: Done!

cazapatamar:
> What is this change for?. the PR has no connected issue

Last week Cesar and Jaime asked for set a limit at 50 for the y axis in this graph due to the final data. It seems the value never get over 50. @Trjegul84 please create the issue and associate.

src/pages/search/drawer/landscape/connectivity/TimelinePAConnectivity.jsx
Trjegul84: You are right. I change it.

cazapatamar (approved):

erikasv (approved):

PR #477, opened 2021-05-25
Feature/download csv files

Trjegul84:
NOTES:
- You must add the library with `yarn add react-json-to-csv`
- In this task the examples of CSV Download Button performed where: **Biomes** and **currentPAConnectivity**.
- File Names of CSV downloaded by user must be standardized. I need help to find out the way of name every file according the area that has been selected. (i.e Biotablero_Biomas_CAR_Quindio.csv)
- In the other tabs, select a place for the download button and later the styles will be fixed.

### DIFF
--- a/src/components/DownloadCSV.jsx
+++ b/src/components/DownloadCSV.jsx
@@ -1,47 +1,37 @@
-import React from 'react';
-import CsvDownload from 'react-json-to-csv';
-import SearchContext from 'pages/search/SearchContext';
 import PropTypes from 'prop-types';
+import React from 'react';
+
 import DownloadIcon from '@material-ui/icons/GetApp';
+import CsvDownload from 'react-json-to-csv';
 
-class DownloadCSV extends React.Component {
-    constructor(props) {
-      super(props);
-      this.state = {
-          buttonlabel: 'Descargar Datos',
-      };
-    }
-
-    render() {
-        const { buttonlabel } = this.state;
-        const { data, filename } = this.props;
-        return (
-          <div className="icondown-container">
-            <CsvDownload
-              data={data}
-              title={buttonlabel}
-              filename={filename}
-              style={{
-                cursor: 'pointer',
-                textDecoration: 'none',
-                background: 'none',
-                border: 'none',
-              }}
-            >
-              <DownloadIcon className="icondown" />
-            </CsvDownload>
-          </div>
-        );
-      }
-    }
+const DownloadCSV = ({ data, filename, buttonTitle }) => (
+  <div className="icondown-container">
+    <CsvDownload
+      data={data}
+      title={buttonTitle}
+      filename={filename}
+      style={{
+        cursor: 'pointer',
+        textDecoration: 'none',
+        background: 'none',
+        border: 'none',
+      }}
+    >
+      <DownloadIcon className="icondown" />
+    </CsvDownload>
+  </div>
+);
 
 DownloadCSV.propTypes = {
     data: PropTypes.array,
     filename: PropTypes.string,
+    buttonTitle: PropTypes.string,
   };
+
 DownloadCSV.defaultProps = {
     data: {},
     filename: '',
+    buttonTitle: 'Descargar Datos',
   };
+
 export default DownloadCSV;
-DownloadCSV.contextType = SearchContext;
--- a/src/components/DownloadCSV.jsx
+++ b/src/components/DownloadCSV.jsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import CsvDownload from 'react-json-to-csv';
+import SearchContext from 'pages/search/SearchContext';
+import PropTypes from 'prop-types';
+import DownloadIcon from '@material-ui/icons/GetApp';
+
+class DownloadCSV extends React.Component {
+    constructor(props) {
+      super(props);
+      this.state = {
+          buttonlabel: 'Descargar Datos',
+      };
+    }
+
+    render() {
+        const { buttonlabel } = this.state;
+        const { data, filename } = this.props;
+        return (
+          <div className="icondown-container">
+            <CsvDownload
+              data={data}
+              title={buttonlabel}
+              filename={filename}
+              style={{
+                cursor: 'pointer',
+                textDecoration: 'none',
+                background: 'none',
+                border: 'none',
+              }}
+            >
+              <DownloadIcon className="icondown" />
+            </CsvDownload>
+          </div>
+        );
+      }
+    }
+
+DownloadCSV.propTypes = {
+    data: PropTypes.array,
+    filename: PropTypes.string,
+  };
+DownloadCSV.defaultProps = {
+    data: {},
+    filename: '',
+  };
+export default DownloadCSV;
+DownloadCSV.contextType = SearchContext;
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "prop-types": "15.6.2",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
+    "react-json-to-csv": "^1.0.4",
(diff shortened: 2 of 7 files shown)


erikasv:
`yarn add` is only to add a new dependency, the command adds it to the package.json file. I our case we must install new dependencies running `yarn install`

erikasv:
For the name, I think we should name it with the `area_type` and `area_id`, these values are already on the URL, so it wouldn't add something strange for the user, for example:  "biotablero_biomas_ea_CRQ.csv"

I think that the word "biotablero" makes it too long for the file name, maybe it's better just "biomas_ea_CRQ.csv", but I also think that would be good to check with Jaime if this name is important, in case we need to add the labels to the context.

src/components/DownloadCSV.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import CsvDownload from 'react-json-to-csv';
+import SearchContext from 'pages/search/SearchContext';
```
erikasv, line 3: This import goes at the end and separated by a blank line. The order is: 

1. Language imports
2. blank line
3. Libraries imports
4. blank line
5. Our modules imports
6. blank lines

erikasv: And I prefer sort the imports alphabetically (but we have never made a agreement with that :P)

src/components/DownloadCSV.jsx
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from 'prop-types';
+import DownloadIcon from '@material-ui/icons/GetApp';
+
+class DownloadCSV extends React.Component {
+    constructor(props) {
+      super(props);
+      this.state = {
+          buttonlabel: 'Descargar Datos',
```
erikasv, line 11: This shouldn't be an state, it could be one of two things:

1. Since the state is never changed, it could be a common variable in a functional component.
2. Or it could be a prop allowing it to be customizable

src/components/DownloadCSV.jsx
```diff
@@ -0,0 +1,47 @@
...
+DownloadCSV.propTypes = {
+    data: PropTypes.array,
+    filename: PropTypes.string,
+  };
+DownloadCSV.defaultProps = {
+    data: {},
+    filename: '',
+  };
+export default DownloadCSV;
+DownloadCSV.contextType = SearchContext;
```
erikasv, line 47: You're not using context

erikasv (changes requested):

Trjegul84:
> `yarn add` is only to add a new dependency, the command adds it to the package.json file. I our case we must install new dependencies running `yarn install`

Please explain me, I don't understand well. When I wrote the component if yarn install is performed it did nothing. Then I perform the command and it works. How is the way of adding this new libraries?.

Trjegul84:
> For the name, I think we should name it with the `area_type` and `area_id`, these values are already on the URL, so it wouldn't add something strange for the user, for example: "biotablero_biomas_ea_CRQ.csv"
> 
> I think that the word "biotablero" makes it too long for the file name, maybe it's better just "biomas_ea_CRQ.csv", but I also think that would be good to check with Jaime if this name is important, in case we need to add the labels to the context.

Please, help me. I don't know how to do it. If I get the context and set  it with `${areaId}_${geofenceId}` the result is like states_63 not the Spanish names.

src/components/DownloadCSV.jsx
Trjegul84: Since I dont find this rules in other components, for example, some add react at the beging others at the end. Then I took the example of Forest.jsx.

src/components/DownloadCSV.jsx
Trjegul84: Component now is functional component.

src/components/DownloadCSV.jsx
Trjegul84: I've delete it.

src/components/DownloadCSV.jsx
erikasv: Yeah, as I said we haven't agreed on the sort thing, but actually the Forest component has it right with the separation, in our case react and prop-types are considered "language" imports.

cazapatamar:
> For the name, I think we should name it with the `area_type` and `area_id`, these values are already on the URL, so it wouldn't add something strange for the user, for example: "biotablero_biomas_ea_CRQ.csv"
> 
> I think that the word "biotablero" makes it too long for the file name, maybe it's better just "biomas_ea_CRQ.csv", but I also think that would be good to check with Jaime if this name is important, in case we need to add the labels to the context.

Agree with @erikasv we should make it simple. As we already use areaType and areaId at URL i think the file name can be composed by these parameters.

cazapatamar:
> > `yarn add` is only to add a new dependency, the command adds it to the package.json file. I our case we must install new dependencies running `yarn install`
> 
> Please explain me, I don't understand well. When I wrote the component if yarn install is performed it did nothing. Then I perform the command and it works. How is the way of adding this new libraries?.

I usually install the new dependency with `yarn install <DEPENDENCY>` and that's all.

cazapatamar:
> > For the name, I think we should name it with the `area_type` and `area_id`, these values are already on the URL, so it wouldn't add something strange for the user, for example: "biotablero_biomas_ea_CRQ.csv"
> > I think that the word "biotablero" makes it too long for the file name, maybe it's better just "biomas_ea_CRQ.csv", but I also think that would be good to check with Jaime if this name is important, in case we need to add the labels to the context.
> 
> Please, help me. I don't know how to do it. If I get the context and set it with `${areaId}_${geofenceId}` the result is like states_63 not the Spanish names.

Yes, it's right. 'states' is the areaType and 63 is the areaId. According to @erikasv suggestion and i agree, we should not use the area name. First, because it is not in the context. Second, the names could be too long in some cases.

Trjegul84:
> > > `yarn add` is only to add a new dependency, the command adds it to the package.json file. I our case we must install new dependencies running `yarn install`
> > 
> > 
> > Please explain me, I don't understand well. When I wrote the component if yarn install is performed it did nothing. Then I perform the command and it works. How is the way of adding this new libraries?.
> 
> I usually install the new dependency with `yarn install <DEPENDENCY>` and that's all.

Yes it should be easy but for me this message is returned when I do that for any library:
```
Unknown Syntax Error: Extraneous positional argument ("anylibrarywhatever").

$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--skip-builds]
```

erikasv:
> ...
> Yes it should be easy but for me this message is returned when I do that for any library:

@Trjegul84 @cazapatamar please read the first part of this documentation: https://classic.yarnpkg.com/en/docs/cli/install/

`yarn install` by itself is to install project's dependencies.

The command @cazapatamar mentions it's to install a missing the dependency in your codebase, that means we could run `yarn install react-json-to-csv` to install that specific dependency in our clones.

`yarn add` is to add a new dependency to the project's package.json (and obviously install it in your copy) : https://classic.yarnpkg.com/en/docs/cli/add

erikasv:
> Please, help me. I don't know how to do it. If I get the context and set it with `${areaId}_${geofenceId}` the result is like states_63 not the Spanish names.

That's exactly my point, not to use names but ids, we should agree in one name template and then ask Jaime how important is this name (explaining that names are not currently accessible from the button)

erikasv:
I think this could be merge now, and after we decide the name this afternoon, we can change it with the implementation of the rest of the buttond. @cazapatamar what do you think?

cazapatamar:
Agarre!

cazapatamar:
> I think this could be merge now, and after we decide the name this afternoon, we can change it with the implementation of the rest of the buttons. @cazapatamar what do you think?

Agree!

erikasv (approved):

PR #479, opened 2021-05-27
Feature/download human footprint

cazapatamar:


erikasv:
lgtm, but i have a question, i started using "fc" instead of "factor_de_compensacion" in the file name for the rest of the charts in compensation factor. should i use the complete name? it would match more with the names used here, but i thought it was too large

cazapatamar:
> lgtm, but i have a question, i started using "fc" instead of "factor_de_compensacion" in the file name for the rest of the charts in compensation factor. should i use the complete name? it would match more with the names used here, but i thought it was too large

Yes,  maybe in some cases could be too long. I will change it using hh instead of the complete name.

Trjegul84:
Then will we use the names in English?

erikasv:
> Then will we use the names in English?

No, hh is in spanish

cazapatamar:
> I mean, the name will be mixed with English (states ea etc) and Spanish (section name). What we chose all English or Spanish?

Good point, i have not argument for or against so what you decide it is ok for me :D

erikasv:
I think states and ea are treated ad id, and shouldn't be translates, but if you want to, we should do it in one place, maybe export the structure from an existing file or create another as we have made in the backend

cazapatamar:
> I think states and ea are treated ad id, and shouldn't be translates, but if you want to, we should do it in one place, maybe export the structure from an existing file or create another as we have made in the backend

According to what @erikasv said i would suggest then let the functionality name in spanish (short) and the areaType in english as it is an id and shouldn't be translated. Something like _bt_hh_states_44.csv_. What do you think? @Trjegul84

Trjegul84:
> > I think states and ea are treated ad id, and shouldn't be translates, but if you want to, we should do it in one place, maybe export the structure from an existing file or create another as we have made in the backend
> 
> According to what @erikasv said i would suggest then let the functionality name in spanish (short) and the areaType in english as it is an id and shouldn't be translated. Something like _bt_hh_states_44.csv_. What do you think? @Trjegul84

I think It is strange the mixed languages, is better is all in English then it would be **bt_hf_states_44.csv**  We may set like this example and later we ask to the thematic group.

cazapatamar:
> > > I think states and ea are treated ad id, and shouldn't be translates, but if you want to, we should do it in one place, maybe export the structure from an existing file or create another as we have made in the backend
> > 
> > 
> > According to what @erikasv said i would suggest then let the functionality name in spanish (short) and the areaType in english as it is an id and shouldn't be translated. Something like _bt_hh_states_44.csv_. What do you think? @Trjegul84
> 
> I think It is strange the mixed languages, is better is all in English then it would be **bt_hf_states_44.csv** We may set like this example and later we ask to the thematic group.

Ok, for now i will let in this way, all in english.

erikasv:
ok, np

erikasv:
But the section names are still in spanish

cazapatamar:
> But the section names are still in spanish

Fixed!

Trjegul84 (approved):

PR #484, opened 2021-05-28
Feature/downstyles

CAScgutierrez:


Trjegul84 (approved):

src/main.css
```diff
@@ -1182,6 +1182,10 @@ overflow: auto;
 	z-index:2;
 }
 
+.contentView * {
+	transition: all 1s ease;
+}
+
```
cazapatamar, line 1188: These lines add an effect on all tooltips. That's expected behaviour? @CAScgutierrez

cazapatamar (commented):
There is a weird effect on tooltips

erikasv (approved):

PR #498, opened 2021-07-07
Feature/draw in map

alegnaaived:
Reviewing with Cesar


erikasv:
I think we should create another component to handle all callbacks and logic related to the polygon, and have only a function to update the final polygon, instead of sending all those functions from Search

cazapatamar:
I agree with the current implementation to draw polygone. Just a couple of question to the future:

1. Should we allow to draw more than one polygone?
2. What should happen when the polygone is confirmed (✓ check button)?

erikasv:
@cazapatamar I'd like to know why do you prefer the current implementation? or you mean only the functional workflow? I really think we should decouple the polygon logic from Search component

cazapatamar:
> @cazapatamar I'd like to know why do you prefer the current implementation? or you mean only the functional workflow? I really think we should decouple the polygon logic from Search component

I meant the functional workflow. I agree to have a new component to handle all the polygone logic in order to decouple this feature and reduce complexity over Search code.

alegnaaived:
> @alegnaaived I agree with the current implementation to draw polygone. Just a couple of question to the future:
> 
> 1. Should we allow to draw more than one polygone?
> 2. What should happen when the polygone is confirmed (✓ check button)?

1. I think no because one polygon will be the parameter to search in our data, but I saw in your screenshot two polygons :/
2. I guess we should send a response with information inside the polygon, but I am not sure what will be the response content. We need to discuss it.

alegnaaived:
> > @alegnaaived I agree with the current implementation to draw polygone. Just a couple of question to the future:
> > 
> > 1. Should we allow to draw more than one polygone?
> > 2. What should happen when the polygone is confirmed (✓ check button)?
> 
> 1. I think no because one polygon will be the parameter to search in our data, but I saw in your screenshot two polygons :/
> 2. We need to keep the polygon in the map and I guess we should send a response with information inside the polygon, but I am not sure what will be the response content. We need to discuss it.

But definitely, we need to keep the polygon drawn in the map.

erikasv:
I still have to make some minor adjustments with Cesar, but this is almost ready.

Do you guys think we should create a separated file for components `AreaAutocomplete` and `Geofences` and another one for `DrawPolygon` ? o just let them inside `Selector.jsx` ? I wasn't sure if it was worth it.

@cazapatamar @Trjegul84 @alegnaaived

cazapatamar:
> I still have to make some minor adjustments with Cesar, but this is almost ready.
> 
> Do you guys think we should create a separated file for components `AreaAutocomplete` and `Geofences` and another one for `DrawPolygon` ? o just let them inside `Selector.jsx` ? I wasn't sure if it was worth it.
> 
> @cazapatamar @Trjegul84 @alegnaaived

I prefer to have separate files for `Geofences` (including `AreaAutocomplete`) and `DrawPolygon` in order to keep simple and readable the Selector component.

cazapatamar (approved):

alegnaaived:
If I select the Check and then press another drawing button, the tooltip keeps there. Do you think it should be another issue to solve in another branch and for the next sprint @erikasv?
image

alegnaaived:
Tooltips for drawing tools could have the same text as inside the accordion. For "Avanzar" tooltip, it says "Enviar polígono", and it is different than the accordion content. Should we validate these messages and texts with César?

image

erikasv:
> If I select the Check and then press another drawing button, the tooltip keeps there. Do you think it should be another issue to solve in another branch and for the next sprint @erikasv?

I spent 2 hours trying to change that, but I couldn't. I didn't find a way to ref some of those sub-menu buttons (the ones from leaflet), or to pass some id to them. If you know a way to do it or wanna try, please.

erikasv:
> Tooltips for drawing tools could have the same text as inside the accordion. For "Avanzar" tooltip, it says "Enviar polígono", and it is different than the accordion content. Should we validate these messages and texts with César?

They are validated with Cesar, also he could change them when he adjust the styles

CAScgutierrez:
> Tooltips for drawing tools could have the same text as inside the accordion. For "Avanzar" tooltip, it says "Enviar polígono", and it is different than the accordion content. Should we validate these messages and texts with César?
> 
Want different texts since is an complex action for user to understand

CAScgutierrez:
> > If I select the Check and then press another drawing button, the tooltip keeps there. Do you think it should be another issue to solve in another branch and for the next sprint @erikasv?
> 
> I spent 2 hours trying to change that, but I couldn't. I didn't find a way to ref some of those sub-menu buttons (the ones from leaflet), or to pass some id to them. If you know a way to do it or wanna try, please.

We talked about this, not the best beahviour but I understand is a hard issue since they are different controls. Ideally people shouldnt be opening those buttons at the same time, but your call if you want to look more into it.

PR #520, opened 2021-09-14
Feature/indicators package

alegnaaived:


packages/indicators/src/cardManager/Card.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import LinkIcon from '../icons/link.svg';
```
erikasv, line 3: I want to know @Trjegul84 and @cazapatamar opinion on this, I didn't like to import svg directly in the polygon functionality to be able to control size and color as props, see [PolygonIcon](https://github.com/PEM-Humboldt/biotablero-frontend/blob/develop/src/pages/search/selector/PolygonIcon.jsx) as an example. Should we keep this strategy for all custom svg icons?

packages/indicators/src/CardManager.jsx
```diff
@@ -0,0 +1,43 @@
...
+  <Masonry options={masonryOptions}>
+    {cardsData.map((card) => (
+      <Card
+        key={card.title}
+        indicator={card.title}
+        title={card.title}
+        goal={card.goal}
+        period={card.period}
+        scale={card.scale}
+        action="./Indicadores"
```
erikasv, line 21: instead of `action` you should use `url` or `href` since it actually is a link

packages/indicators/src/icons/link.svg
```diff
@@ -0,0 +1,18 @@
...
+        0-7.071l4.998-4.998c.975-.976 2.256-1.464 3.536-1.464 1.279 0 2.56.488 3.535 
+        1.464.493.493.861 1.063 1.105 1.672l-.787.784zm-5.703.147c.178.643.521 1.25 
+        1.026 1.756 1.562 1.563 4.096 1.561 5.656 0l4.999-4.998c1.563-1.562 1.563-4.095 
+        0-5.657-1.562-1.562-4.095-1.563-5.657 0l-3.841 3.841-.333-.009c-.404 
+        0-.802.04-1.189.117l4.656-4.656c.975-.976 2.256-1.464 3.536-1.464 1.279 0 
+        2.56.488 3.535 1.464 1.951 1.951 1.951 5.119 0 7.071l-4.999 4.998c-.975.976-2.255 
+        1.464-3.535 1.464-1.28 0-2.56-.488-3.535-1.464-.494-.495-.863-1.067-1.107-1.678l.788-.785z" 
+        fill="#e84a5f"/>
+    </defs>
+</svg>
```
erikasv, line 18: new line missing

packages/indicators/src/CardManager.jsx
erikasv: And it's not static

packages/indicators/src/data/selectorData.js
```diff
@@ -0,0 +1,60 @@
+const cardsData = [
+  {
+    id: '1',
```
erikasv, line 3: missing link in data

erikasv (changes requested):
Important note:
**A lot of commits aren't in the message format we agreed on the workflow document.**

Other adjustments
* The package must be able to run independently, at this moment when I run `yarn workspace indicators run dev` throws this error: `Error: No transformers found for packages/indicators/src/icons/link.svg`
* When I try to run the package inside BioTablero it throws this error when it detects changes: `[!] Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)`
* We agreed to follow the [fractal structure](https://hackernoon.com/fractal-a-react-app-structure-for-infinite-scale-4dab943092af), so `CardManager` should be inside another folder named `app`. Inside this folder also should be `data` and `cardManager`
* When I ran `yarn install` the `yarn.lock` file was updated, so it means it's outdated in the branch
* All the code related with indicators in the main project should be deleted.
* `Indicators.jsx` must be renamed to `Indicator.jsx`

alegnaaived:
Changes done in commits messages according to [workflow document](https://docs.google.com/document/d/1TGc0AX6aIhxKNArgszvij6KRoHZ0uYSzuTMfRvDDO-E/edit#)

> Important note:
> **A lot of commits aren't in the message format we agreed on the workflow document.**
> 
Other adjustments (In progress)

src/main.css
```diff
@@ -1320,6 +1320,15 @@ overflow: auto;
   height: 65px;
 }
 
+.Indicadores {
```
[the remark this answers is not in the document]
erikasv: My mistake, the classes are seted dynamically

erikasv:
@alegnaaived Why did you renamed the package?

packages/indicator/src/App.jsx
```diff
@@ -0,0 +1,39 @@
...
+            </button>{' '}
+            Filtros de búsqueda{' '}
+          </h3>
+          {openFilter && (
+            <div>
+              <h4>
+                <button type="button" onClick={() => setOpenFilter(!openFilter)}>
+                  x
+                </button>{' '}
+                (Prueba de apertura del menú)
```
erikasv, line 26: This button isn't to close the menu, is to clear all applied filters. You can just delete it because it's out of scope for this issue

packages/indicator/src/app/CardManager.jsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const CardManager = ({ cardsData }) => (
+  <>
+    <div className="countD">
+      {cardsData ? `${cardsData.length} indicadores` : 'No hay indicadores'}
+    </div>
+    <div className="countDIcon">
+      <GetAppOutlinedIcon color="#e84a5f" />
+    </div>
```
erikasv, line 19: It looks funny, maybe works better if you put them inside same div?


Also, check SaveAlt or SystemUpdateAlt icons, they look more similar to the one Cesar put in mockups, check with him which one fits better

erikasv: Also I think this section fits better in the App component, as you had it before, so CardManager only care about organizing cards. But if you want to keep it here, move the cards data extraction to this component, instead of a prop

packages/indicator/src/app/cardManager/Card.jsx
```diff
@@ -0,0 +1,48 @@
...
+          onKeyDown={() => {}}
+          role="button"
+          title="Abrir indicador"
+          tabIndex={0}
+        >
+          +
+        </div>
+      </span>
+    </h1>
+    <h2>{period}</h2>
```
erikasv, line 25: This prop is optional, you should render it only if it's present (to avoid posible spaces created by margin and padding later)

packages/indicator/src/app/data/selectorData.js
```diff
@@ -0,0 +1,68 @@
...
+    externalLink: './Indicadores',
+  },
+  {
+    id: '8',
+    title: '',
+    goal: '',
+    period: '',
+    scale: '',
+    externalLink: './Indicadores',
+  },
```
erikasv, line 65: What is the purpose of this case?

packages/indicator/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-prettier": "^3.3.1",
+    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-react-hooks": "^4.2.0",
+    "parcel": "^2.0.0-rc.0",
+    "prettier": "^2.3.2",
+    "rollup": "^2.56.3",
+    "rollup-plugin-css-only": "^3.1.0",
+    "rollup-plugin-react-svg": "^3.0.3"
```
erikasv, line 34: What is this dependency for? I removed it and it works fine

erikasv (changes requested):
1. I don't understand why you changed the package name, the instruction in the previous review was to rename the component (in pages) from the main project, but the package is independent. The change makes terribly difficult to identify the changes you introduced. Although if every one agree it's better to keep it singular, update the documentation so it can be a reliable example.

2. When the console is open I noticed a strange behavior, then added more test data and it replicated:

packages/indicator/src/app/data/selectorData.js
alegnaaived: Showing default styles, but I can remove it

packages/indicators/src/app/data/selectorData.js
```diff
@@ -0,0 +1,68 @@
...
+    externalLink: './Indicadores',
+  },
+  {
+    id: '8',
+    title: '',
+    goal: '',
+    period: '',
+    scale: '',
+    externalLink: './Indicadores',
+  },
```
cazapatamar, line 65: Also agree with deleting this case

cazapatamar (changes requested):

packages/indicators/src/App.jsx
```diff
@@ -0,0 +1,39 @@
...
+          {openFilter && (
+            <div>
+              <h4>
+                <button type="button" onClick={() => setOpenFilter(!openFilter)}>
+                  x
+                </button>{' '}
+                (Prueba de apertura del menú)
+              </h4>
+            </div>
+          )}
```
cazapatamar, line 29: I would delete this div as it is only a test. According to the task scope you should only let a side empty panel, where there will be filters.

cazapatamar (changes requested):

cazapatamar:
> @alegnaaived Why did you renamed the package?

I think `Indicators` was a suitable name to the package. I also didn't understand this change.

alegnaaived:
> @alegnaaived Why did you renamed the package?

Due to it was a misunderstanding, I changed it again.

packages/indicators/src/app/CardManager.jsx
```diff
@@ -0,0 +1,42 @@
...
+        lastUpdate={card.lastUpdate}
+        scale={card.scale}
+        externalLink={card.externalLink}
+      />
+    ))}
+  </Masonry>
+);
+
+CardManager.propTypes = {
+  cardsData: PropTypes.arrayOf(
```
erikasv, line 28: instead of redefining the shape set `arrayOf` to be the Card proptypes, in that component they are well defined

erikasv (changes requested):

packages/indicators/src/app/cardManager/Card.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import LinkIcon from './card/URLIcon';
```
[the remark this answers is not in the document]
alegnaaived: Done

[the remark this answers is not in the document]
alegnaaived: Done

packages/indicators/src/App.jsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react';
+
+import './main.css';
+import CardManager from './app/CardManager';
+import cardsData from './app/data/selectorData';
+import DownloadIcon from './app/cardManager/card/DownloadIcon';
```
erikasv, line 6: `DownloadIcon` should be in `app` folder (not `card`). Please review the fractal post.

packages/indicators/src/App.jsx
```diff
@@ -0,0 +1,54 @@
...
+import OpenIcon from './app/cardManager/card/OpenIcon';
```
erikasv, line 7: `OpenIcon` should be in a new folder `src/components`

packages/indicators/src/App.jsx
```diff
@@ -0,0 +1,54 @@
...
+import CloseIcon from './app/cardManager/card/CloseIcon';
```
erikasv, line 8: This is also in the wrong location

packages/indicators/src/app/CardManager.jsx
```diff
@@ -0,0 +1,31 @@
...
+        lastUpdate={card.lastUpdate}
+        scale={card.scale}
+        externalLink={card.externalLink}
+      />
+    ))}
+  </Masonry>
+);
+
+CardManager.propTypes = {
+  cardsData: PropTypes.arrayOf(Card).isRequired,
```
erikasv, line 28: I think the correct way is to reference the Card propTypes, beacuse you are nor receiving the component itself, but its props. [See this](https://stackoverflow.com/questions/59283221/how-to-define-proptypes-for-a-component-that-only-passes-props-to-a-child/59283558#59283558).

But I just noticed you have 2 differences: `key` and `indicator` are Card props, but in this component are `id` and `title`, so you could change some of the props to make it easier.

erikasv (changes requested):
Also run `format` script in the package

packages/indicators/src/App.jsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react';
+
+import './main.css';
+import CardManager from './app/CardManager';
+import cardsData from './app/data/selectorData';
+import DownloadIcon from './app/DownloadIcon';
+import OpenIcon from './components/OpenIcon';
+import CloseIcon from './components/CloseIcon';
```
erikasv, line 8: Why did you put this inside `components` folder? I only see it used by app.

packages/indicators/src/app/CardManager.jsx
```diff
@@ -0,0 +1,29 @@
...
+        lastUpdate={card.lastUpdate}
+        scale={card.scale}
+        externalLink={card.externalLink}
+      />
+    ))}
+  </Masonry>
+);
+
+CardManager.propTypes = {
+  ...Card.propTypes,
```
erikasv, line 26: CardManager receives one prop named cardsData which is an array of objects. Not as many props as the Card component

erikasv (changes requested):

erikasv (approved):

cazapatamar (approved):

PR #523, opened 2021-09-23
Fix/message error data

cazapatamar:


src/pages/search/drawer/species/richness/NumberOfSpecies.jsx
```diff
@@ -254,6 +261,12 @@ class NumberOfSpecies extends React.Component {
           Haga click en cada barra para visualizar su mapa,
           que corresponden a los datos inferidos.
         </h3>
+        {showErrorMessage && (
+          <div className="disclaimer">
+            La riqueza inferida del área de consulta supera la de la región biótica en algunos
```
erikasv, line 266: According to this message it only applies for inferred, so the validation shouldn't be made over observed data? or the message should be a little more general?

cazapatamar: Yes, according to Cesar, the message should only includes inferred because observed case never occurs.

I just deleted the observed validation.

erikasv (approved):

PR #561, opened 2021-11-05
Feature/bug states and back icon

Trjegul84:
NOTE:
In Accordion. If collapsed prop in component Didmount of an accordion comes in true a condition rejects the tab and search for next accordion. By default the first tab is expanded because when the first tab comes with collapsed in false it will be the default tab. In all tabs by default will expand the first. 
To avoid pa and listAreas tabs expanded at the same time (because both come with collapsed in false) when user returns with backIcon collapsed must be set in true by default in `SearchAreas.jsx`. In this way all tabs comes in collapsed true, when user leaves the section all tabs will be collpased avoiding the first tab to be expanded.

### DIFF
--- a/src/pages/Search.jsx
+++ b/src/pages/Search.jsx
@@ -908,7 +908,7 @@ class Search extends Component {
         () => {
           const { history } = this.props;
           const { selectedAreaType, selectedArea } = this.state;
-          if (selectedAreaType) {
+          if (selectedAreaType && selectedArea) {
             history.push(`?area_type=${selectedAreaType.id}&area_id=${selectedArea.id || selectedArea.name}`);
             setHeaderNames(selectedAreaType.name, selectedArea.name);
           }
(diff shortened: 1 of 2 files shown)


src/pages/Search.jsx
```diff
@@ -908,8 +908,10 @@ class Search extends Component {
         () => {
           const { history } = this.props;
           const { selectedAreaType, selectedArea } = this.state;
-          history.push(`?area_type=${selectedAreaType.id}&area_id=${selectedArea.id || selectedArea.name}`);
-          setHeaderNames(selectedAreaType.name, selectedArea.name);
+          if (selectedAreaType) {
```
erikasv, line 911: Seeing this again, I think it would be better to add `selectedArea` to the if too to "curarnos en salud" :sweat_smile: .

erikasv (changes requested):

src/pages/Search.jsx
Trjegul84: ok

erikasv (approved):

PR #570, opened 2021-11-17
Fix cancel raster request

erikasv:
I couldn't make it less coupled, but I think it's a little simpler now since it doesn't use the state to store the last active request.

Now all the active requests sources are stored, not only the last one, and when they finish the source is deleted. But the key point is not to block `switchLayer`, the problem was the `await` statement for the map legend values.


Trjegul84 (approved):

cazapatamar (approved):

alegnaaived (commented):
I like it too

PR #572, opened 2021-11-19
Change prop in mapviewer to receive an array of urls instead of a string

cazapatamar:


cazapatamar:
This PR is the result of the spike about multiple raster:

1. This code is to enable MapViewer to receive an array instead of a single string with urls of rasters
2. We can let this branch as a reference to the moment when the rasters are ready of ´Perdida y persistencia´ section or we can merge now. Let me know your opinions.
3. The bounds will remain the same even if we render multiple rasters as they corrensponds to geofence and not to raster

src/components/MapViewer.jsx
```diff
@@ -218,7 +220,7 @@ MapViewer.propTypes = {
   layers: PropTypes.object.isRequired,
   // eslint-disable-next-line react/no-unused-prop-types
   layerError: PropTypes.bool,
-  rasterLayer: PropTypes.string,
+  rasterLayers: PropTypes.array,
```
erikasv, line 223: use [arrayOf](https://reactjs.org/docs/typechecking-with-proptypes.html) and specify type

src/pages/Search.jsx
```diff
@@ -772,11 +772,11 @@ class Search extends Component {
       Promise.all(promises).then(([res, legendValues]) => {
         this.activeRequests.delete(layerType);
         if (res !== 'request canceled') {
-          const rasterUrl = `data:${res.headers['content-type']};base64, ${Buffer.from(res.data, 'binary').toString('base64')}`;
+          const rasterUrls = [`data:${res.headers['content-type']};base64, ${Buffer.from(res.data, 'binary').toString('base64')}`];
```
erikasv, line 775: we'll have to change this logic again T-T (but don't do anything right now)

erikasv (changes requested):

erikasv (approved):

PR #578, opened 2021-11-26
Coverages layer

cazapatamar:


cazapatamar:
The issue with data was solved using sql script to modify area field. As data in layer anyway is not matching exactly with data in graphs so we have decided with Jaime to avoid showing area in ha inside the tooltip so it is only shown the name of category (Natural, Secundario, Transformada, Nubes). The above, due to different ways of thematic team to compute this data.

Please test response time as it is taking too long for me. Have in account the shape is too heavy and is very detailed.

We can discuss these two points tomorrow.

erikasv:
This feature broke the rasters map, since they take the bounds from the geofence layer, I suggest yo to add the call to `switchLayer` with `geofence` previous loading rasters to avoid this.

cazapatamar:
> This feature broke the rasters map, since they take the bounds from the geofence layer, I suggest you to add the call to `switchLayer` with `geofence` previous loading rasters to avoid this.

Fixed

erikasv (approved):

PR #604, opened 2022-02-22
Material migration v5

erikasv:
@cazapatamar @Trjegul84 @alegnaaived Please run the app and check all modules and sections, I did it myself, but just in case

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,5 +17,8 @@
         "paths": ["src"]
       }
     }
+  },
+  "env": {
+    "browser": true
   }
 }
--- a/package.json
+++ b/package.json
@@ -9,9 +9,12 @@
     "lint": "eslint src --ext .js,.jsx"
   },
   "dependencies": {
-    "@material-ui/core": "^4.0.0",
-    "@material-ui/icons": "^4.9.1",
-    "@material-ui/lab": "^4.0.0-alpha.57",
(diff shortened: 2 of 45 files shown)


cazapatamar:
This error shows up by opening the subzones selector. We might need a key prop either in Autocomplete or TextField at SearchAreas component.

imagen

Trjegul84 (changes requested):
Check when you go to Compensaciones >> Agregar nuevo proyecto an error is displayed
Screenshot_20220222_161913

erikasv:
@cazapatamar Agree, names are repeated, but ids are unique, so I made a change to use ids as keys for the autocomplete component

erikasv:
@Trjegul84 Solved, it had to do with [this](https://github.com/mui/material-ui/issues/29690#issuecomment-971673478)

cazapatamar:
Could you please add 

> @cazapatamar Agree, names are repeated, but ids are unique, so I made a change to use ids as keys for the autocomplete component

Could you please add this solution to the modal `NUEVO PROYECTO` field `Estado del proyecto`

imagen

erikasv:
@cazapatamar project status don't have an id, I did a workaround with the value. but the problem could reappear.

cazapatamar (approved):

Trjegul84 (approved):

PR #606, opened 2022-02-24
Upgrade nivo to 0.79.1

cazapatamar:


erikasv (changes requested):
I see a strange square whit the tooltip in all bars (Ecosystems, CF, HF, Integrity, Connectivity), can you get rid of it?:
image
image

Although we didn't manage to solve completely the overflow, it's now more hidden:
image

Also tooltips in the pie chart seem have lost the animation, if you compare them with production there is something off in their movement in this new version, maybe a prop was added to control the animation?

cazapatamar:
@alegnaaived Please check the erika's comment and let me know if you have the same behavior after deleting browser cache.

cazapatamar:
> I see a strange square whit the tooltip in all bars (Ecosystems, CF, HF, Integrity, Connectivity), can you get rid of it?: image image
> 
> Although we didn't manage to solve completely the overflow, it's now more hidden: image
> 
> Also tooltips in the pie chart seem have lost the animation, if you compare them with production there is something off in their movement in this new version, maybe a prop was added to control the animation?

> I see a strange square whit the tooltip in all bars (Ecosystems, CF, HF, Integrity, Connectivity), can you get rid of it?: image image
> 
> Although we didn't manage to solve completely the overflow, it's now more hidden: image
> 
> Also tooltips in the pie chart seem have lost the animation, if you compare them with production there is something off in their movement in this new version, maybe a prop was added to control the animation?

Please check the animation. The small bar graphs are not able to have this effect because we need to set absolute position in css to avoid the tooltip get out of the panel (Ecosystems tab).

erikasv (approved):
You were right. The strange square and the tooltip movement in the pie chart are ok (I now even see the animation in the chart, maybe we should check on Cesar about this later).

And as I was writing this, I saw the new commit :)

erikasv:
> Please check the animation. The small bar graphs are not able to have this effect because we need to set absolute position in css to avoid the tooltip get out of the panel (Ecosystems tab).

Yeah, it was only in the pie chart that seemed off.

cazapatamar:
> 

Yeah, it was already validated with Cesar. In addition, The last commit includes a new effect over the pie sections that i consider looks nice.

PR #607, opened 2022-02-24
Change layers control

erikasv:


cazapatamar:
If i select CORPOBOYACA (avoiding charge completly the coverage layer ) and open the `ESPECIES` tab the map keeps loading and don't show any layer.

imagen

For CARDER it works fine.

Trjegul84:
I can not replicate the same behavior as @cazapatamar mention with CORPOBOYACA works.  I think something happen in any section / tab with any unit , when you enter to one tab and layer is not completely load and you switch very fast to other section the new layer takes to long to load. Another thing I see sometimes is the delay to show the loading bars (CARGANDO) over the map.

erikasv:
Sorry, I couldn't replicate the behavior @cazapatamar reported.

About @Trjegul84 comment, I see the delay, but can't find the source of it, I think there are too many points were layers are shut off, I don't know if it's such a big deal and we could treat it when we make the refactor of `switchLayer`, what do you think?

Trjegul84:
Yes I agree. Maybe it may require more analysis & testing.

Trjegul84 (approved):

cazapatamar:
> Sorry, I couldn't replicate the behavior @cazapatamar reported.
> 
> About @Trjegul84 comment, I see the delay, but can't find the source of it, I think there are too many points were layers are shut off, I don't know if it's such a big deal and we could treat it when we make the refactor of `switchLayer`, what do you think?

Maybe in my case it was a network issue

cazapatamar (approved):

PR #608, opened 2022-02-25
Fix dots chart tooltip and action

erikasv:


src/components/charts/DotsGraph.jsx
```diff
@@ -150,6 +150,7 @@ export default withTooltip(({
             backgroundColor: 'rgba(0,0,0,0.9)',
             padding: 12,
             lineHeight: '1.5',
+            position: 'absolute',
```
cazapatamar, line 153: Please set `fontSize: 14` in tooltip style as text looks a little bit bigger than production

cazapatamar (changes requested):

src/components/charts/DotsGraph.jsx
Trjegul84: Yeah, I already check it. It seams big font is the only issue.

erikasv: Done

cazapatamar (approved):

Trjegul84 (approved):

PR #613, opened 2022-03-08
Add condition to identify when component is rendered from Search or Compensation module

cazapatamar:


erikasv:
Related issue? I think this corresponds to #611 which also needs to be in release, so it also should be created from release.

If it's needed later in develop, we'll just merge release into develop

PR #615, opened 2022-03-08
Enable download strategies button again

erikasv:


Trjegul84 (changes requested):
Take a  look at the error "Warning: Encountered two children with the same key, \`7-302-2802-CORPOGUAJIRA\`. Keys should be unique so ..." is displayed when you select any sub basin zone in selector list.
For example in the case of >> EN LICENCIAMIENTO >> El Paso La Jagua Sep >>
GEB ZONA NORTE

Screenshot_20220309_080157

erikasv:
> Take a look at the error "Warning: Encountered two children with the same key, `7-302-2802-CORPOGUAJIRA`. Keys should be unique so ..." is displayed when you select any sub basin zone in selector list. For example in the case of >> EN LICENCIAMIENTO >> El Paso La Jagua Sep >> GEB ZONA NORTE

The problem is caused by the previously saved strategies, somehow there are repeated strategies for the same combination of biome,  subzone and ea. It's out of scope of this task, and I don't even know if we can fix it without risking the data.

Trjegul84 (approved):

PR #616, opened 2022-03-08
Feature/update text portfolio

alegnaaived:


src/pages/Portfolio.jsx
```diff
@@ -1,5 +1,5 @@
-import Item from 'pages/portfolio/Item';
 import React from 'react';
+import Item from './portfolio/Item';
```
erikasv, line 2: don't use relative imports

erikasv (changes requested):

erikasv (approved):

PR #617, opened 2022-03-10
Coverage update data

cazapatamar:
New colors already validated with Cesar

This PR depends on branch in the backend which name is `feature/coverageColors`

### DIFF
--- a/src/pages/Search.jsx
+++ b/src/pages/Search.jsx
@@ -393,7 +393,8 @@ class Search extends Component {
     switch (chartType) {
       case 'coverage':
         this.highlightRaster(`${chartType}-${selectedKey}`);
-      break;
+        break;
+      // Current progress of the refactor
       case 'paramo':
         this.shutOffLayer('wetland');
         this.shutOffLayer('dryForest');
@@ -738,6 +739,10 @@ class Search extends Component {
     let mapLegend = null;
 
     switch (layerType) {
+      case 'coverages':
+        this.setSectionLayers(layerType);
+        return;
+      // Current progress of the refactor
       case undefined:
       case null:
         this.cancelActiveRequests();
@@ -918,9 +923,6 @@ class Search extends Component {
           };
         });
         break;
-      case 'coverages':
-        this.setSectionLayers(layerType);
-        return;
       default:
         if (/SciHfPA-*/.test(layerType)) {
           const [, sci, hf] = layerType.match(/SciHfPA-(\w+)-(\w+)/);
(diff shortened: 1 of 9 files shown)


src/pages/Search.jsx
```diff
@@ -48,6 +48,11 @@ const tooltipLabel = {
   X: 'Sin clasificar / Nubes',
 };
 
+const rasterOpacity = {
+  default: 0.7,
+  selected: 1,
+};
+
```
erikasv, line 55: I don't like this object here, it can vary from section to section the value of the "default opacity", it's better to have the value set for each section or since it's needed from various functions, save it in the rasterUrls state.

I think we could always assume the highlighted raster section will have opacity 1.

src/pages/Search.jsx
```diff
@@ -335,13 +340,48 @@ class Search extends Component {
...
     layers[layerName].layer.resetStyle(feature);
     feature.closePopup();
   }
 
+  /**
+   * Highlight specific raster on the map
+   *
+   * @param {String} layerName Raster layer name
+   */
+  highlightRaster = (layerName) => {
```
erikasv, line 355: According to the comparison, this param is in reality the layer id, not name

erikasv (changes requested):

Trjegul84 (dismissed):

src/pages/Search.jsx
cazapatamar: Done

src/pages/Search.jsx
cazapatamar: Add default opacity to `activeLayer` state and delete `rasterOpacity` object that was out of the class. I think this solution allows to set a different opacity to each case at `setSectionLayers` function.

src/pages/Search.jsx
```diff
@@ -503,6 +542,172 @@ class Search extends Component {
...
+
+  setSectionLayers = (sectionName) => {
+    this.setState({ loadingLayer: true, layerError: false });
+    this.shutOffLayer();
+
+    let baseLayerId = null;
+    const shapeLayerIds = [];
+    let rasterLayerIds = [];
+    const newActiveLayer = { id: sectionName };
+    newActiveLayer.defaultOpacity = 1;
```
erikasv, line 647: Add `defaultOpacity` key to the object declared in the previous line

erikasv (changes requested):

erikasv (approved):

PR #620, opened 2022-03-16
Refactor richness layers

cazapatamar:


src/pages/Search.jsx
```diff
@@ -668,7 +685,64 @@ class Search extends Component {
...
+                  matchColor('richnessGaps')('legend-to'),
+                ],
+              };
+              return newState;
+            });
+          },
+        };
+        break;
       default:
+        if (/numberOfSpecies*/.test(sectionName)) {
```
erikasv, line 716: What do you think if we avoid the switch and only use `if - else` statements? I'm liking that option more than this combination.

cazapatamar: I agree. Already done!

erikasv (approved):

PR #632, opened 2022-04-07
Strategic ecosystem update data

cazapatamar:


src/pages/Search.jsx
```diff
@@ -697,6 +710,14 @@ class Search extends Component {
       rasterLayerIds = ['coverage-N', 'coverage-S', 'coverage-T'];
       newActiveLayer.name = 'Coberturas';
       newActiveLayer.defaultOpacity = 0.7;
+    } else if (/seCoverages*/.test(sectionName)) {
+      let seType = 'paramo';
+      const selected = sectionName.match(/seCoverages-(\w+)/);
+      if (selected) [, seType] = selected;
+      baseLayerId = 'geofence';
+      rasterLayerIds = [`seCoverage-${seType}-N`, `seCoverage-${seType}-S`, `seCoverage-${seType}-T`];
+      newActiveLayer.name = `Número de especies - ${SELabel(seType)}`;
```
erikasv, line 719: ¿Número de especies?

src/pages/search/drawer/StrategicEcosystems.jsx
```diff
@@ -37,8 +37,8 @@ class StrategicEcosystems extends React.Component {
       coverage: [],
       protectedAreas: [],
       PAArea: 0,
-      strategicEcosistems: [],
-      SEArea: 0,
+      SEAreas: [],
+      SETotalArea: 0,
```
erikasv, line 41: Good call, way better names

src/pages/Search.jsx
cazapatamar: Fixed!

erikasv (changes requested):
Great job organizing the components' variables and props. :clap: 

You're missing the handling of the case when there is an empty layer for just one or two coverage types (not all of them), check the case of arauca for example (id: 81) the area for coverage `S` is so small that the layer is empty and the front is showing the `CAPA NO DISPONBLE` error.

Trjegul84 (commented):
Something happen when you first expand information in Páramo and then if you expand another SE for example Tropical Dry forest it  sends an error like: `TypeError: np is undefined`

Screenshot_20220407_145352

Take a look of this case:
area_type=ea&area_id=CORPOCHIVOR

cazapatamar:
> Something happen when you first expand information in Páramo and then if you expand another SE for example Tropical Dry forest it sends an error like: `TypeError: np is undefined`
> 
> Screenshot_20220407_145352
> 
> Take a look of this case: area_type=ea&area_id=CORPOCHIVOR

This bug is actually in develop, it seems it is related to distribution of protected areas in selected SE. Apparently we are missing to set the section in the bar graph asociated to category `No protected` (Grey color) after adjustments over protected area endpoints.

I think it is better to solve it in another task. What do you think? @erikasv @Trjegul84 

imagen

Trjegul84:
Ok, I agree it could be handled in a fix bug task.

cazapatamar:
> Great job organizing the components' variables and props. clap
> 
> You're missing the handling of the case when there is an empty layer for just one or two coverage types (not all of them), check the case of arauca for example (id: 81) the area for coverage `S` is so small that the layer is empty and the front is showing the `CAPA NO DISPONBLE` error.

Solved!

erikasv:
> I think it is better to solve it in another task. What do you think? @erikasv @Trjegul84

Sorry for the delay, yeah, this should be solved with the new calculation of percentages in protected areas, right now this only happens with environmental authorities because of the issue with the new query, after we solve the issue backend-294 it will happen with all area types.

erikasv:
Sorry @cazapatamar, the error Angela mentioned also happens in states and in that case is not related with the changes done in the protected areas query.

The problem is related when there are no protected areas at all in the strategic ecosystem, so, although the logic of that function (`setPAValues`) will need to be changed, it is better to add that control for when there are no protected areas. (just add that condition to the firs if :P)

src/pages/Search.jsx
```diff
@@ -796,11 +821,12 @@ class Search extends Component {
       });
     }
 
+    const isMultipleLayers = rasterLayerIds.length > 1;
```
erikasv, line 824: With this solution you can't know if all the layers failed, it think it would be better to change the condition of this `if (rasterLayers.includes('canceled'))` to return the error only if all rasters failed

Trjegul84:
@cazapatamar @erikasv  Please tell me if the following behavior is desirable: If you expand any strategic ecosystem then you can no longer back to see the Coverage. I mean you can't return to Coverage bar and look its layer because click action is kind of disabled. In the same page the only way to see again the Coverage Layer is refreshing it.

cazapatamar:
> Something happen when you first expand information in Páramo and then if you expand another SE for example Tropical Dry forest it sends an error like: `TypeError: np is undefined`
> 
> Screenshot_20220407_145352
> 
> Take a look of this case: area_type=ea&area_id=CORPOCHIVOR

Fixed!

src/pages/Search.jsx
cazapatamar: Fixed!

src/pages/Search.jsx
```diff
@@ -601,6 +606,8 @@ class Search extends Component {
   /**
    * Request a raster layer from the backend
    * @param {String} layerName name of the layer to request
+   * @param {Boolean} isMultipleLayers indicate whether or not the layer is part of a set of
+   * simultaneous layers in order to avoid return conection error if one fails
```
erikasv, line 610: delete obsolete doc

erikasv (approved):

PR #640, opened 2022-04-21
Feature/new pa endpoints

cazapatamar:


src/pages/search/utils/transformData.js
```diff
@@ -0,0 +1,71 @@
+export const transformPAValues = (rawData, totalArea) => {
```
erikasv, line 1: Maybe we could add here the function `processData` that is currently in Ecosystems component ? Or do you think is out of scope for this file?

erikasv (commented):
Great job organizing EcosystemDetails :tada:

src/pages/search/utils/transformData.js
cazapatamar: Yes, agree. The idea is to use this file to centralize the functions that transform data to be passed to graph components specially those that are used from diferent components.

cazapatamar: Done!

cazapatamar:
When there is not data of PA for an speciic SE, a message `Sin áreas protegdas` is shown. It is already validated with Cesar.

PR #687, opened 2022-07-12
Landscape texts

cazapatamar:


src/pages/search/drawer/landscape/humanFootprint/TimelineFootprint.jsx
```diff
@@ -201,7 +207,7 @@ class TimelineFootprint extends React.Component {
         {(
           showInfoGraph && (
             <ShortInfo
-              description={info}
+              description={texts.hfTimeline.info}
```
erikasv, line 210: I noticed sometimes the text takes a little to load, but while it's loading `undefined` is shown, I think it would be better to show `loading...` while the texts load.

And that should be the case in every `ShortInfo` section, since those texts are open by default

erikasv (changes requested):

src/pages/search/drawer/landscape/humanFootprint/TimelineFootprint.jsx
erikasv: Actually, would be even better to handle this in the `ShortInfo` component

erikasv (approved):

PR #690, opened 2022-07-22
Feature/texts into state

Trjegul84:
Please check if something extra in `NumberOfSpecies.jsx` is needed.


cazapatamar:
I think it is not necessary any change over NOS componente as it already includes texts as state of component.

cazapatamar (approved):

PR #694, opened 2022-08-03
Migrate ForestIntegrity component

erikasv:
Some adjustments on other components were necessary


src/pages/search/drawer/landscape/forest/ForestIntegrity.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+  componentDidMount() {
+    this.mounted = true;
+    const { areaId, geofenceId, switchLayer } = this
+      .context as SearchContextValues;
+
+    switchLayer("forestIntegrity");
+
+    SearchAPI.requestSCIHF(areaId, geofenceId)
+      .then((res: Array<SCIHF>) => {
```
Trjegul84, line 126: In TimelinePAConnectivity this res was not highlighted for any type needed. In what cases this type definition for res will be needed?

src/utils/searchAPI.ts
```diff
@@ -14,12 +17,27 @@ class SearchAPI {
   static requestSCIHF(
     areaType: string,
     areaId: string | number
-  ): Promise<SCIHF[]> {
+  ): Promise<Array<SCIHF>> {
```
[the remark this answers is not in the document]
Trjegul84: Are the types needed also in requestTimelinePAConnectivity and also in all requests of restAPI.js ?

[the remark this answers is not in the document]
erikasv: Yes, we talk this last week, we're going to migrate requests from restAPI.js to SearchAPI.ts so we can type them progressively

src/pages/search/drawer/landscape/forest/ForestIntegrity.tsx
erikasv: Something must be wrong then, the type is required unless the other variables this response interacts with are any

src/pages/search/types/texts.ts
```diff
@@ -0,0 +1,6 @@
+export interface TextObject {
```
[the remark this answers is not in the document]
erikasv: I thought it was necessary only for set .js files types. I'll read about it

cazapatamar (approved):

src/pages/search/types/texts.ts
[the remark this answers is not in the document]
erikasv: According to the [documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) and [this answer](https://stackoverflow.com/a/50464124) I think `*.d.ts` files are appropriate to define common js modules (and they must have the same name as the module they are defining), so I wouldn't create the files with the shared types with this extension.

Last week we decided to create a file for each "*big accordion*" to hold shared definitions (that means types for the state will be in each component file and not in this folder `types`). I still like this decision, but if you have another suggestion we can discuss it.

Also @cazapatamar , @ManuelStardust what do you think?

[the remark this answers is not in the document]
cazapatamar: Sorry i haven't read this comment.

I think having a ´*.d.ts´ file for every object retrieved from the dabatase could increase the complexity of the code and it would leads to have many independent files. I agree to create a file for each accordion and there hold all share definitions.

Trjegul84 (approved):

src/pages/search/drawer/landscape/forest/ForestIntegrity.tsx
```diff
@@ -0,0 +1,282 @@
...
+      showInfoGraph: !prevState.showInfoGraph,
+    }));
+  };
+
+  render() {
+    const {
+      showInfoGraph,
+      SciHfCats,
+      ProtectedAreas,
+      isDataNull,
```
cazapatamar, line 194: This state is not used

src/utils/searchAPI.ts
```diff
@@ -14,12 +17,27 @@ class SearchAPI {
   static requestSCIHF(
     areaType: string,
     areaId: string | number
```
cazapatamar, line 19: It is necessary to put a comma at the end of this line?

cazapatamar (changes requested):

src/pages/search/drawer/landscape/forest/ForestIntegrity.tsx
erikasv: You're right, solved.

src/utils/searchAPI.ts
erikasv: I thought so, but prettier remove is :woman_shrugging:

cazapatamar (approved):

PR #701, opened 2022-08-05
Migrate current pa connectivity component to typescript

cazapatamar:


src/utils/searchAPI.ts
```diff
@@ -23,6 +24,44 @@ class SearchAPI {
...
+   * @param {String} areaType area type id, f.e. "ea", "states"
+   * @param {Number | String} areaId area id to request, f.e. "CRQ", 24
+   *
+   * @return {Promise<Object>} Array of objects with data of the protected areas
+   */
+  static requestDPC(
+    areaType: string,
+    areaId: string | number,
+    paNumber: number
+  ) {
```
erikasv, line 59: Please add the return type (you imprted it but isn't being used xD)

src/utils/searchAPI.ts
```diff
@@ -1,5 +1,6 @@
 import axios, { CancelToken } from "axios";
```
erikasv, line 1: Please remoe CancelToken, we'll add it when we need it

erikasv (changes requested):

src/utils/searchAPI.ts
cazapatamar: Done!

src/utils/searchAPI.ts
cazapatamar: Fixed!

erikasv (approved):

PR #711, opened 2022-08-11
TS: migrate forest loss component

erikasv:
also migrate chart associated with it


src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
```diff
@@ -1,57 +1,72 @@
...
-    const {
-      areaId,
-      geofenceId,
-      switchLayer,
-    } = this.context;
-
-    const getPersistenceValue = (data) => {
-      const periodData = data ? data.find((item) => item.id === LATEST_PERIOD).data : null;
-      const persistenceData = periodData ? periodData.find((item) => item.key === 'persistencia') : null;
+    const { areaId, geofenceId, switchLayer } = this
```
Trjegul84, line 54: For me the following lines are a bit ticky!!

Trjegul84 (approved):

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
erikasv: are you referring to the destructuring assignment?

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
```diff
@@ -1,57 +1,72 @@
...
+
+    const getPersistenceValue = (rawData: Array<ForestLP>) => {
+      const periodData = rawData.find(
+        (item) => item.id === LATEST_PERIOD
+      )?.data;
+      const persistenceData = periodData
+        ? periodData.find((item) => item.key === "persistencia")
+        : null;
       return persistenceData ? persistenceData.area : 0;
     };
```
erikasv, line 65: Guys I'll wait for your thoughts on this syntax.

In my opinion it's easy enough once explained, and I like it more than ternary operators or if statements.

Trjegul84: Thanks @erikasv for the explanation. I agree, the first option results in more elegant code than ternary.

cazapatamar: Yes, i think once we start to use it will become easy. For me the explanation was enought to undertand the difference between using ternary operator compared to this implementation. Also agree to leave this implementation as it is.

PR #717, opened 2022-08-17
Migrate CurrentSEPAConnectivity to Typescript

ManuelStardust:
The component was converted to typescript, Prettier was executed on the component. The correct functioning of the component was verified.

PR #719, opened 2022-08-18
Migrate to typescript compensation factor component

cazapatamar:


src/pages/search/drawer/landscape/CompensationFactor.tsx
```diff
@@ -121,14 +164,14 @@ class CompensationFactor extends React.Component {
             texts: { ...prevState.texts, [item]: {} },
```
erikasv, line 164: I think is because of the `[item]` this line doesn't throw the ts error thrown in other components, but as for the type definition it's better to just not do anything in case of failure (in no other place are the texts modified, so the structure would be kept)

src/pages/search/drawer/landscape/CompensationFactor.tsx
```diff
@@ -164,29 +192,27 @@ class CompensationFactor extends React.Component {
...
-        bioticunits: bioticUnitsMess,
+        bioticUnits: bioticUnitsMess,
       },
       texts,
     } = this.state;
-    const {
-      areaId,
-      geofenceId,
-    } = this.context;
+    const { areaId, geofenceId, handlerClickOnGraph } = this
```
erikasv, line 199: Why did you added `handlerClickOnGraph` ? it's not used

erikasv (changes requested):

src/pages/search/drawer/landscape/CompensationFactor.tsx
cazapatamar: Do you mean ..... `[item]: { info: "", cons: "", meto: "", quote: "" }`?

src/pages/search/drawer/landscape/CompensationFactor.tsx
cazapatamar: Fixed!

src/pages/search/drawer/landscape/CompensationFactor.tsx
erikasv: Yeap, see in other components that the catch for the test simply doesn't do anything

cazapatamar: Ok i will fix it.

erikasv (approved):
Pleas wait until the PR [717](https://github.com/PEM-Humboldt/biotablero-frontend/pull/717) is merged

PR #721, opened 2022-08-18
Migrate strategic ecosystems to typescript

Trjegul84:
Please help me with these questions:

1. In `SearchAPI` I place the ecosystems section below the human footprint section, is there any order established for adding the sections there?

1. Also the `PAData` interface defined has the label as string because it has a lot of options. Should all the options for the label property be defined in a constant in ecosystems.ts  ?

1.  In this task, Should I migrate to typescript the `Ecosystems.jsx `component also?  

1. The endpoint related with `requestSEPAByGeofence` returns something like  
```
[
  {
    "area": 447657.310088262,
    "type": "Parque Nacional Natural"
  }
]
```
  But I set the response as an array of `PAData` as far I can understand the return of `transformPAValues` function and you can see the structure of that object is different. Should I need different typescript interface for this response only with `area` and `type` or it is ok setting the return as  `Promise<Array<PAData>>`


src/pages/search/types/ecosystems.ts
```diff
@@ -0,0 +1,27 @@
...
+] as const;
+export const coverageType = ["N", "S", "T", "X"] as const;
+
+export interface SEValues {
+  area: number;
+  percentage?: number;
+  type: string;
+}
+
+export interface CoverageData {
```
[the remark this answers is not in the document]
erikasv: Also, once you move the specific component use definitions to that file, the coverageLabels should be moved to the component too

src/pages/search/types/ecosystems.ts
```diff
@@ -0,0 +1,27 @@
...
+  area: number;
+  type: typeof coverageType[number];
+  percentage: number;
+  key: typeof coverageType[number];
+  label: typeof coverageLabels[number];
+}
+
+export interface PAData {
```
erikasv, line 22: According to the documentation, the endpoint returns an array of objects with 'area' and 'type'. Please check that the documentation is up to date and make the same fix o the two previous interfaces

erikasv (changes requested):

erikasv:
>     1. In `SearchAPI` I place the ecosystems section below the human footprint section, is there any order established for adding the sections there?
There is no order, just grouping, it's ok there

>     2. Also the `PAData` interface defined has the label as string because it has a lot of options. Should all the options for the label property be defined in a constant in ecosystems.ts  ?
I didn't notice those mentioned options, maybe explain them in the daily

>     3. In this task, Should I migrate to typescript the `Ecosystems.jsx `component also?
No, actually I was thinking only in one of the two components you migrated xD

>     4. The endpoint related with `requestSEPAByGeofence` returns something like
> ...
> But I set the response as an array of `PAData` as far I can understand the return of `transformPAValues` function and you can see the structure of that object is different. Should I need different typescript interface for this response only with `area` and `type` or it is ok setting the return as `Promise<Array<PAData>>`
Seethr comments in the code

src/pages/search/drawer/strategicEcosystems/EcosystemsBox.tsx
```diff
@@ -0,0 +1,84 @@
...
+import GraphLoader from "components/charts/GraphLoader";
+import formatNumber from "utils/format";
+import matchColor from "utils/matchColor";
+import { transformSEValues } from "pages/search/utils/transformData";
+import { SEValues } from "pages/search/types/ecosystems";
+
+interface EcosystemsBoxProps {
+  SETotalArea: number;
+  SEAreas: Array<SEValues>;
+  setActiveSE: (type: string) => {};
```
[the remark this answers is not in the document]
Trjegul84: done!

src/pages/search/drawer/strategicEcosystems/EcosystemsBox.tsx
```diff
@@ -0,0 +1,84 @@
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
+import PropTypes from "prop-types";
+import React, { Component } from "react";
+import SearchContext from "pages/search/SearchContext";
```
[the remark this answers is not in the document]
Trjegul84: removed!

src/utils/searchAPI.ts
```diff
@@ -220,6 +221,40 @@ class SearchAPI {
...
+  static requestSEPAByGeofence(
+    areaType: string,
+    areaId: string | number,
+    seType: string
+  ): Promise<Array<SEPAData>> {
+    return SearchAPI.makeGetRequest(
+      `/pa/se?areaType=${areaType}&areaId=${areaId}&seType=${seType}`
+    );
+  }
+
```
erikasv, line 257: These functions should be under the ecosystems section, not in the middle of footprint

src/utils/restAPI.js
```diff
@@ -1,4 +1,4 @@
-import axios, { CancelToken } from 'axios';
+import axios, { CancelToken } from "axios";
```
erikasv, line 1: Undo the format changes in this file so we can see the relevant changes

erikasv (changes requested):

src/utils/restAPI.js
Trjegul84: done!

erikasv (approved):

PR #761, opened 2022-09-21
Rename Graph Components

Trjegul84:
I propose names using a more readable `adjective-noun` structure for the name of graphs, for example:
`StackedLines` instead of `MultiLines` any way they are multiple lines if they are stacked.

`LargeStackedBar` to add large and stacked adjectives before the name of the Nivo component we are using i.e Bar (ResponsiveBar). Also `SmallStackedBar` is used instead of  SmallBarStackGraph.

Since multiple small stacked bar graphic is represented in `MultiSmallBarStackGraph` then the name is changed to `MultiSmallStackedBar`

To represent only one bar, I think it is not needed to point out that it is single so `MultiSmallSingleBarGraph` could be renamed to `MultiSmallBar` only with **Bar without stacked or single. Although, it is intended to represent a graphic with multiple small bars another better option could be `SmallBars` to avoid the multi prefix and use the plural. The same for MultiSmallStackedBar another option could be SmallStackedBars.

Finally, I decided to change `MultiSmallSingleBarGraph` to `SingleSmallBars`, to imply a graph with single small bars in the sense of a group of individual bars is simpler, readable, and more understandable.

### DIFF
--- a/src/pages/search/drawer/Ecosystems.jsx
+++ b/src/pages/search/drawer/Ecosystems.jsx
@@ -17,7 +17,7 @@ import formatNumber from 'utils/format';
 import matchColor from 'utils/matchColor';
 import RestAPI from 'utils/restAPI';
 import SearchAPI from "utils/searchAPI";
-import SmallBarStackGraph from 'pages/search/shared_components/charts/SmallBarStackGraph';
+import SmallStackedBar from 'pages/search/shared_components/charts/SmallStackedBar';
 
 /**
  * Calculate percentage for a given value according to total
@@ -292,7 +292,7 @@ class Ecosystems extends React.Component {
           </h6>
           <div className="graficaeco">
             <div className="svgPointer">
-              <SmallBarStackGraph
+              <SmallStackedBar
                 message={cov}
                 data={coverage}
                 units="ha"
@@ -336,7 +336,7 @@ class Ecosystems extends React.Component {
             <h6>
               Distribución por áreas protegidas:
             </h6>
-            <SmallBarStackGraph
+            <SmallStackedBar
               message={pa}
               data={PAAreas}
               units="ha"
--- a/src/pages/search/drawer/ecosystems/EcosystemsBox.tsx
+++ b/src/pages/search/drawer/ecosystems/EcosystemsBox.tsx
@@ -7,7 +7,7 @@ import matchColor from "utils/matchColor";
 
 import { transformSEValues } from "pages/search/drawer/ecosystems/transformData";
 import { SEPADataExt } from "pages/search/types/ecosystems";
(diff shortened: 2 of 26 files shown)


erikasv:
I'm not convinced with `StackedLines`, but if @cazapatamar and @ManuelStardust agree let's go with it. I like the rest of the names.

ManuelStardust:
I understand that these types of graphs are time series, perhaps include the word "series".

> I'm not convinced with `StackedLines`, but if @cazapatamar and @ManuelStardust agree let's go with it. I like the rest of the names.

cazapatamar:
Sorry for the late response.

@Trjegul84 According to names you propose initially, i suggest some changes in order to standardize the way we call these components:

TimelineFootprint
StackedLines -> Lines

ForestLossPersistence
MultiSmallStackedBar -> SmallStackedBars

CurrentPAConnectivity (DPC)
SingleSmallBars -> SmallBars

NumberOfSpecies
SingleBullet -> Bullet

Below i explain why:

- I dont see necessary the word "Stacked" at "StackedLines". It has a meaning in Bar grahps but not here. Only Lines represent well the graph.
- As you did in others components, we can omit the word "Multi" in order to keep the name simple. I think it is a good idea to use "s" at the end to highlight that component has multiple lines, bars, etc.
- Also, i think we can omit using the word "Single" for the same reason. I mean, use singluar and plural by naming the components. instead of using additional adjectives.

Please let me know your opinion.

erikasv:
I see Camilo's point, the idea is avoid `single` and `multi` and use singular and plural names, to summarize and clarify all names, the current names would be:

- LargeStackedBar -> example: current footprint
- SmallStackedBar -> example: ecosystems
- SmallStackedBars -> example: Loss and Persistence
- SmallBars -> example: connectivity - DPC 
- Pie -> example: Integrity
- Bullet -> example: number of species
- Lines -> example: historic footline

I would agree with those names

cazapatamar:
> > MultiSmallStackedBar -> SmallStackedBars
> 
> @cazapatamar there is a problem with the name MultiSmallStackedBar ( you suggest SmallStackedBars) it is used in ForestLost is used for this group of bars: Screenshot_20220922_105849
> 
> And also we have now SmallStackedBar (before SmallBarStackGraph ) used in ecosystems, ecosystem box, ecosystems details, and ForestIntegrity. The name could be easily confused with the other.

Yes, what i propose is to change MultiSmallStackedBar by SmallStackedBars. Used at Forest LP section.

I like better using plural and singular (SmallStackedBar vs SmallStackedBars) even if the names could be similar.

It is just and opinion but if you consider still using multi and single it is also fine.

cazapatamar (approved):

PR #768, opened 2022-10-07
Adjust Portfolios by Target Graph

ManuelStardust:
- Rename SmallStackedBars to MultiSmallBars
- Update ForestLossPersistence
- Adjust Portfolios by Target data from API to Graph
- Add Portfolios by Target Color Palette

### DIFF
--- a/src/pages/search/drawer/portfolios/conservationAreas/Targets.tsx
+++ b/src/pages/search/drawer/portfolios/conservationAreas/Targets.tsx
@@ -145,11 +145,12 @@ class Targets extends React.Component<Props, State> {
             colors={matchColor("caTargets")}
             units="%"
             onClickHandler={() => {}}
-            height={300}
+            height={500}
             selectedIndexValue="WCMC"
             groupMode="grouped"
             toolTipValue="percentage"
             innerPadding={0.5}
+            marginLeft={145}
           />
         </div>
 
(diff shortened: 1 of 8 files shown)


src/pages/search/drawer/portfolios/conservationAreas/Targets.tsx
```diff
@@ -148,13 +139,17 @@ class Targets extends React.Component<Props, State> {
...
-            colors={matchColor("dpc")}
-            labelX="dPC"
-            units="ha"
+            colors={matchColor("caTargets")}
+            units="%"
             onClickHandler={() => {}}
+            height={300}
+            selectedIndexValue="WCMC"
+            groupMode="grouped"
+            toolTipValue="percentage"
```
erikasv, line 151: I believe we'll have to show more than one field, in this case is possible we need to show the value and the percentage. But let's leave that for another issue after we merge the small bars components.

erikasv (commented):
I see everything ok, but please make the backend changes before merging this one, so we can see more than 2 groups of bars

erikasv (changes requested):
Some names are being cut, please check if you can add some style or parameter to fix it, but verify that doesn't break the other charts
image

erikasv (approved):

erikasv (approved):

PR #773, opened 2022-10-14
Merge SmallBars and MultiSmallBars

ManuelStardust:
- Merge MultiSmallBars into SmallBars 
- Migrate Portfolios By Target
- Migrate Forest Loss And Persistence
- Adapt CurrentPAConnectivity


erikasv (changes requested):
You left too many properties from nivo to be passed to our component. We need to have some basic chart, that implies some default values. I suggest you return to the original one used by LossAndPersistence and let the properties related with style (for example axis, margin, pading, etc) with the values back then as the default ones.

The transformation you are doing in the controllers must be done in the chart component, that transformation is too confusing to manage all data, the previous sprint we discussed with Camilo an approximation to the structure SmallBars should receive, it's closer to what was before. The pending considerations were about what properties in the array of objects keep and or if it was necessary any other.

Don't hard code values (colors) even if they are going to be ignored, and if they are going to be ignored, they shouldn't be in the data.

I don't think it's a good idea to pass the whole html tooltip to the component, I understand there is a difference between the values shown, but we must figure some other way to handle that.

We lost the category highlight in the map for the LossAndPersistence interaction.

You're also missing the removal of MultiSmallBars.

I suggest next time you please be a little more specific in the daily meetings, for example, name a thing that hold you back, so we can all agree or discuss earlier some decisions.

PR #780, opened 2022-10-21
Portfolios legends

erikasv:
I'm sorry for the big changes.

I made some improvements to the way the data is requested, also keep the backend order for the targets and renamed legend components to be clearer.

And finally I had to adjust the margin bottom of SmallBars, so.. changed the margin prop


cazapatamar (approved):

ManuelStardust (approved):
Everything works.

PR #783, opened 2022-11-01
Portfolios CA layers adjusts

ManuelStardust:
- Add cancelToken in search API
- Adjust regex pattern in getRasterLayer


cazapatamar:
The layer don't disappear when close the accodion

image

It seems is happenig with all layers

image

erikasv (approved):

cazapatamar (approved):

PR #792, opened 2022-11-15
Post Polygon Submit View

ManuelStardust:


src/pages/Search.jsx
```diff
@@ -1379,6 +1381,15 @@ class Search extends Component {
     this.setState({ [state]: false });
   };
 
+  /**
+   * Set search type value
+   *
+   * @param {String} state state value to set the search type
+   */
+  setSearchType(searchType) {
```
erikasv, line 1389: change this to arrow function

src/pages/Search.jsx
```diff
@@ -1470,9 +1484,11 @@ class Search extends Component {
               mapTitle={mapTitle}
               drawPolygonEnabled={drawPolygonEnabled}
               loadPolygonInfo={this.loadPolygonInfo}
+              setSearchType={this.setSearchType.bind(this)}
```
erikasv, line 1487: remove the `bind` part (with arrow function it's not necessary)

src/pages/search/mapViewer/DrawControl.jsx
```diff
@@ -243,3 +246,5 @@ DrawControl.defaultProps = {
 };
 
 export default DrawControl;
+
+DrawControl.contextType = SearchContext;
```
erikasv, line 250: add newline at the end

src/pages/Search.jsx
erikasv: Also these 2 props aren't necessary, both calls can be done inside `loadPolygonInfo` in this same component

src/pages/search/drawer/Ecosystems.jsx
```diff
@@ -62,100 +62,103 @@ class Ecosystems extends React.Component {
...
-      });
-
-    SearchAPI.requestProtectedAreas(areaId, geofenceId)
-      .then((res) => {
-        if (this.mounted) {
-          if (Array.isArray(res) && res[0]) {
-            const PATotalArea = res.map((i) => i.area).reduce((prev, next) => prev + next);
-            const PAAreas = transformPAValues(res, generalArea);
+    
+    if(searchType!=="polygon"){
```
erikasv, line 69: Change the condition to return if it's "polygon" and avoid the extra indentation level

erikasv: Also, since we had to deactivate eslint, pay more attention to code style, in this case add a space after the `if` and between the comparation symbol

src/pages/search/drawer/Ecosystems.jsx
```diff
@@ -242,151 +245,161 @@ class Ecosystems extends React.Component {
...
-          </IconTooltip>
-          {infoShown.has('coverage') && (
+    if(searchType==="polygon"){
+      return (
+        <div className="graphcard">
+            <h2>Gráficas en construcción</h2>
+            <p>Pronto más información</p>
+          </div>
+      );
+    }else{
```
erikasv, line 257: this else is unnecessary, remove it and avoid the extra indentation

src/pages/search/drawer/Ecosystems.jsx
```diff
@@ -242,151 +245,161 @@ class Ecosystems extends React.Component {
...
-            </h4>
-          </button>
-          <IconTooltip title="Interpretación">
-            <InfoIcon
-              className={`downSpecial${infoShown.has('coverage') ? ' activeBox' : ''}`}
-              onClick={() => this.toggleInfo('coverage')}
-            />
-          </IconTooltip>
-          {infoShown.has('coverage') && (
+    if(searchType==="polygon"){
```
erikasv, line 250: please fix style (see comment above)

src/pages/search/drawer/Species.jsx
```diff
@@ -23,25 +23,29 @@ class Species extends React.Component {
...
-    switch (areaId) {
-      case 'states':
-        if (geofenceId !== '88') {
-          selected = ['richness', 'functionalDiversity'];
-        }
-        break;
-      case 'ea':
-        if (geofenceId !== 'CORALINA') {
+
+    if(searchType!=="selected"){
```
erikasv, line 29: "selected" isn't a valid value for this variable, this is working only because `areaId` is null and it's entering the default case

erikasv: Also I don't see the if necessary, @cazapatamar I think we could handle this inside the default case of the switch, later if it's necessary something for the polygon we can change it. What do you think?

Also this would apply for all Tabs (Landscape and Portfolios), and just make sure when setting `searchType` to polygon to also set areaId and geofenceId to null

src/pages/search/drawer/landscape/Forest.tsx
```diff
@@ -1,12 +1,16 @@
...
 import ForestIntegrity from "pages/search/drawer/landscape/forest/ForestIntegrity";
 import ForestLossPersistence from "pages/search/drawer/landscape/forest/ForestLossPersistence";
+import SearchContext from "pages/search/SearchContext";
 import { accordionComponent, componentProps } from "pages/search/types/ui";
 
 const Forest: React.FC<componentProps> = (props) => {
   const { handleAccordionChange, openTab } = props;
 
-  const componentsArray: Array<accordionComponent> = [
+  const { searchType } = useContext(SearchContext);
```
erikasv, line 11: In the Search module we don't use hooks, It's ok for me to have a combination of classes and hooks (we have hooks in other modules already), but I don't know if it can be confusing, @cazapatamar how would you feel about it?

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
```diff
@@ -47,34 +47,60 @@ class ForestLossPersistence extends React.Component<Props, State> {
...
-      .getForestLPTexts("forestLP")
-      .then((res) => {
-        if (this.mounted) {
-          this.setState({ texts: { forestLP: res } });
-        }
-      })
-      .catch(() => {});
+    if (searchType === "polygon") {
+      this.flpController
+        .getForestLPData(areaId, geofenceId, LATEST_PERIOD)
```
erikasv, line 55: hardcode an area, don't leave `areaId` and `geofenceId` since they are null

erikasv: Also, I think this logic should go inside the controller, I don't like this giant if. `getForestLPData` should receive everything and return the data the way it can be stored in the state.

For example, it could receive `areaId`, `geofenceId`, `searchType` and the polygon.

src/pages/search/SearchContext.tsx
```diff
@@ -3,6 +3,7 @@ import React from "react";
 export interface SearchContextValues {
   areaId: string;
   geofenceId: string | number;
+  searchType: "selection" | "polygon";
```
erikasv, line 6: I think you need to store the polygon in the context too, or it would be necessary to pass it down as prop or to manage all requests from Search, which is a bad idea

erikasv (changes requested):

src/pages/search/drawer/Species.jsx
cazapatamar: Yes, i think we can delete that unuseful condition and use de default case to handle when there is a selected geofence or a polygon. Likely in the near future we can use a similar `if` when we are clearer about what to do in each case.

src/pages/search/SearchContext.tsx
cazapatamar, line 6: @erikasv are you ok with `selection` name for this option? I think it sounds too general and i like better something like `geofence`, `defGeofence`, `definedGeofence` or similar. Anyway if you agree with `selection` it is ok for me.

On the other hand, i don't like `polygon` either considering in the future (long future) we can implement the option to upload a polygon and specific rules associated, so this name in the context could be too general. I suggest a name like `drawPolygon` or similar.

src/pages/search/drawer/landscape/Forest.tsx
erikasv: I see, We can discuss it tomorrow, but I think that if we want to use hooks in the module, starting with the big accordion components (like this case: `Forest`) would be a good point

src/pages/search/SearchContext.tsx
erikasv: Yes, I agree with your suggestions, I like more `definedArea` (more than "geofence" since we started deprecating the word a while ago xD) and `drawPolygon`

cazapatamar: definedArea sounds good

ManuelStardust: Done

ManuelStardust: Done, I added a context var called polygonRequest

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
ManuelStardust: Done

src/pages/search/drawer/Species.jsx
ManuelStardust: Done

src/pages/Search.jsx
ManuelStardust: This function was removed, the change in searchType is done directly in _loadPolygonInfo_

src/pages/Search.jsx
ManuelStardust: Done.

src/pages/search/mapViewer/DrawControl.jsx
ManuelStardust: Done

src/pages/search/drawer/Ecosystems.jsx
ManuelStardust: Done

src/pages/search/drawer/Ecosystems.jsx
ManuelStardust: Done

src/pages/search/drawer/Ecosystems.jsx
ManuelStardust: Done

src/pages/search/SearchContext.tsx
```diff
@@ -11,6 +14,8 @@ export interface SearchContextValues {
 const SearchContext = React.createContext<SearchContextValues>({
   areaId: "",
   geofenceId: "",
+  searchType: "definedArea",
+  polygonRequest: {} as Polygon,
```
erikasv, line 18: what is the purpose of `as Polygon` ?

src/pages/search/SearchContext.tsx
```diff
@@ -1,8 +1,11 @@
 import React from "react";
+import { Polygon } from "pages/search/types/drawer";
 
 export interface SearchContextValues {
   areaId: string;
   geofenceId: string | number;
+  searchType: "definedArea" | "drawPolygon";
+  polygonRequest: {};
```
erikasv, line 8: `polygonRequest` is always an empty object?

src/pages/search/drawer/landscape/forest/ForestLossPersistenceController.ts
```diff
@@ -35,9 +36,15 @@ export class ForestLossPersistenceController {
   getForestLPData = (
     areaType: string,
     areaId: string | number,
-    latestPeriod: string
-  ): Promise<ForestLPData> =>
-    SearchAPI.requestForestLP(areaType, areaId)
+    latestPeriod: string,
+    searchType: "definedArea" | "drawPolygon",
+    polygonRequest: {}
```
erikasv, line 41: Why would this always be an empty object?

src/pages/search/SearchContext.tsx
ManuelStardust: It's because initially when I declared the empty object it generated an error, and the way I found to solve it was this. However, checking again I think that I can initialize the variable in the following way: _polygonRequest: { latLngs: [] }_, what do you think?

ManuelStardust: Fixed

src/pages/search/SearchContext.tsx
ManuelStardust: Fixed

src/pages/Search.jsx
```diff
@@ -1321,7 +1324,12 @@ class Search extends Component {
           layer: L.polygon(polygon.latLngs, { fitBounds: true }),
         },
       },
+      selectedAreaTypeId: "",
+      selectedAreaId: "",
```
erikasv, line 1328: It's better accept null for these 2 variables in the context, I see in the constructor and in `handlerBackButton` they are being set to null

src/pages/search/drawer/landscape/forest/ForestLossPersistenceController.ts
```diff
@@ -29,15 +30,23 @@ export class ForestLossPersistenceController {
    * @param {String} areaType area type
    * @param {String | Number} areaId area id
    * @param {String} latestPeriod string with range of years for latest period
+   * @param {String} searchType string to identifyy the type of search
```
erikasv, line 33: typo in identifyy

src/pages/search/drawer/landscape/forest/ForestLossPersistenceController.ts
```diff
@@ -29,15 +30,23 @@ export class ForestLossPersistenceController {
...
+   * @param {Object | null} polygonRequest object with range of years for latest period
```
erikasv, line 34: I don't understand this description what do you mean with `with range of years for latest period`?

ManuelStardust: Sorry, it was my mistake. Fixed.

src/pages/Search.jsx
ManuelStardust: Done

erikasv (approved):

PR #794, opened 2022-11-18
Add target name to portfolios map logic

cazapatamar:
Add target name to rendering layer logic in order to be able to send that info to backend

### DIFF
--- a/src/pages/search/drawer/portfolios/conservationAreas/Targets.tsx
+++ b/src/pages/search/drawer/portfolios/conservationAreas/Targets.tsx
@@ -86,7 +86,10 @@ class Targets extends React.Component<Props, State> {
                       );
                     if (portfoliosIds && portfoliosIds.size > 0) {
                       switchLayer(
-                        ["portfoliosCA", [...portfoliosIds].join("-")].join("-")
+                        [
+                          `portfoliosCA|${target.target_name}`,
+                          [...portfoliosIds].join(","),
+                        ].join("|")
                       );
                       this.setState({ selectedPortfolios: portfoliosIds });
                     }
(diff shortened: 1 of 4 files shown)


cazapatamar:
@erikasv @ManuelStardust Ready!

src/utils/searchAPI.ts
```diff
@@ -527,12 +527,13 @@ class SearchAPI {
...
     areaType: string,
     areaId: string,
+    targetName: string,
     portfolioId: number
   ): { request: Promise<Object>; source: CancelTokenSource } {
     const source = axios.CancelToken.source();
     return {
       request: SearchAPI.makeGetRequest(
-        `portfolios-ca/portfolios/layer?areaType=${areaType}&areaId=${areaId}&portfolioId=${portfolioId}`,
+        `portfolios-ca/portfolios/layer?areaType=${areaType}&areaId=${areaId}&targetName=${targetName}&portfolioId=${portfolioId}`,
```
erikasv, line 536: You didn't made any modification to the backend to do something with this param. And as long as I remember the portfolio layer is completely independent from the targets

src/pages/search/drawer/portfolios/conservationAreas/Targets.tsx
```diff
@@ -86,7 +86,10 @@ class Targets extends React.Component<Props, State> {
                       );
                     if (portfoliosIds && portfoliosIds.size > 0) {
                       switchLayer(
-                        ["portfoliosCA", [...portfoliosIds].join("-")].join("-")
+                        [
+                          `portfoliosCA|${target.target_name}`,
+                          [...portfoliosIds].join(","),
+                        ].join("|")
```
erikasv, line 92: This is too big and confusing to keep handling it with Arrays, I think it's better to construct the String directly with a template

src/utils/searchAPI.ts
cazapatamar: You are right, i misunderstood, we have never have talk about relationship between targets and portfolios layer.

src/pages/search/drawer/portfolios/conservationAreas/Targets.tsx
cazapatamar: Done!

erikasv (approved):

PR #797, opened 2022-11-24
Draw polygon map

cazapatamar:


src/pages/search/drawer/landscape/forest/ForestLossPersistenceController.ts
```diff
@@ -31,7 +31,7 @@ export class ForestLossPersistenceController {
    * @param {String | Number} areaId area id
    * @param {String} latestPeriod string with range of years for latest period
    * @param {String} searchType string to identify the type of search
-   * @param {Object | null} polygonRequest object with the coordinates array
+   * @param {Array | null} polygon Coordinates of polygon
```
erikasv, line 34: Polygon type as define in drawer is an object, not an Array. 

But in the JSDoc task meeting we decided to adopt [TSDoc](https://tsdoc.org/), which requires deleting the types from the doc comment. Manuel is doing a POC, but I think we'll stick to it. So, if you want, instead of correcting the type, you could just delete types from this function doc comment.

src/pages/search/types/drawer.ts
```diff
@@ -1,7 +1,9 @@
...
+
 export interface geofenceDetails {
   total_area: string;
 }
 
-export type Polygon = {
-  latLngs: Array<{ lat: number; lng: number }>;
-};
+export interface Polygon {
+  copordinates: LatLngExpression[][];
```
erikasv, line 8: typo in coordinates

erikasv: Also, this prop should be just `LatLngExpression[]` since it already considers `[number, number]` as an accepted type

src/pages/search/drawer/landscape/forest/ForestLossPersistenceController.ts
cazapatamar: Types deleted

src/pages/search/types/drawer.ts
cazapatamar: Fixed

cazapatamar:
@erikasv please check the last [commit](https://github.com/PEM-Humboldt/biotablero-frontend/pull/797/commits/4098b115b874b28af9192c641622362cbd687144) where i use the already defined type `Polygon` in the state in the `MapViewer` component.

I had overlooked this state in `MapViewer` and actually is the same type already defined.

erikasv (approved):

erikasv:
> I had overlooked this state in `MapViewer` and actually is the same type already defined.

I didn't notice either :sweat_smile:

PR #817, opened 2023-03-24
Error messages when no connection to any of backends

cazapatamar:


src/pages/Search.jsx
```diff
@@ -145,16 +149,40 @@ class Search extends Component {
...
+      }
+    }));
+  }
+
+  /**
+   * Get the current list of scripts at BIAB backend. It is used to check connection to backend.
+   */
+  checkPolygonConn = () => {
+    biabAPI.requestScriptList()
+      .then(() => {})
```
erikasv, line 172: this is unnecessary, just leave the catch

erikasv (changes requested):

src/pages/Search.jsx
cazapatamar: Done!

erikasv (approved):

PR #824, opened 2023-03-31
Message for selector when there is no connection or loading

cazapatamar:


src/pages/Search.jsx
```diff
@@ -1484,8 +1495,8 @@ class Search extends Component {
         <Modal
           aria-labelledby="simple-modal-title"
           aria-describedby="simple-modal-description"
-          open={connErrors.defAreas && connErrors.polygon}
-          onClose={this.handleCloseModal('connError')}
+          open={messages.defAreas === 'conn-error' && messages.polygon === 'conn-error' && openErrorModal}
```
erikasv, line 1498: the first two conditions are redundant, just with the `openErrorModal` is enough

erikasv (changes requested):

src/pages/Search.jsx
cazapatamar: Done!

erikasv (approved):

PR #830, opened 2023-05-08
Calculate and display total area for query by polygon

ManuelStardust:


src/pages/Search.jsx
```diff
@@ -44,6 +44,8 @@ class Search extends Component {
       rasterUrls: [],
       searchType: "definedArea",
       polygon: null,
+      polygonArea: 0,
+      setPolygonArea: this.setPolygonArea,
```
erikasv, line 48: this shouldn't be in the state

src/pages/search/SearchContext.tsx
```diff
@@ -7,6 +7,8 @@ export interface SearchContextValues {
   searchType: "definedArea" | "drawPolygon";
   polygon: Polygon | null;
   polygonFolder: string;
+  polygonArea: number;
+  setPolygonArea(value?: number): void;
```
erikasv, line 11: I don't wee why allow the parameter to be optional

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
```diff
@@ -39,6 +40,7 @@ class ForestLossPersistence extends React.Component<Props, State> {
       forestLP: [],
       message: "loading",
       forestPersistenceValue: 0,
+      forestLPArea: 0,
```
erikasv, line 43: why is it optional?

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
```diff
@@ -20,6 +20,7 @@ interface State {
   forestLP: Array<ForestLPExt>;
   message: wrapperMessage;
   forestPersistenceValue: number;
+  forestLPArea?: number;
```
erikasv, line 23: why is it optional?

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
```diff
@@ -74,11 +77,13 @@ class ForestLossPersistence extends React.Component<Props, State> {
           this.setState({
             forestLP: data.forestLP,
             forestPersistenceValue: data.forestPersistenceValue,
+            forestLPArea: data.forestLPArea,
             message: null,
           });
         }
         if (searchType === "drawPolygon") {
           switchLayer(`forestLP-${LATEST_PERIOD}`);
+          setPolygonArea(this.state.forestLPArea);
```
erikasv, line 86: I think it would be better to use `data.forestLPArea` directly

src/pages/search/drawer/landscape/forest/ForestLossPersistenceController.ts
```diff
@@ -13,6 +13,7 @@ import { Polygon } from "pages/search/types/drawer";
 interface ForestLPData {
   forestLP: Array<ForestLPExt>;
   forestPersistenceValue: number;
+  forestLPArea?: number;
```
erikasv, line 16: why is it optional

ManuelStardust: because this value is only used for the case of query by polygon, in the case of query by defined area it is not used. I think the other option is to return forestLPArea as zero for the defined area case.

src/pages/Search.jsx
ManuelStardust: Solved

src/pages/search/SearchContext.tsx
ManuelStardust: Solved

src/pages/search/drawer/landscape/forest/ForestLossPersistence.tsx
ManuelStardust: Solved

PR #834, opened 2023-05-24
Disable layer loading modal when response fails

ManuelStardust:


erikasv:
I don't think this is a good solution, is making use of `switchLayer` for something isn't intended for.

I think what we discussed is to load again the polygon from switchLayer (besides this would align with the next task)

cazapatamar:
The solution you implemented is one of the options I initially proposed. However, based on our discussion, the possible solution aligns with what Erika mentioned in the comment above.

ManuelStardust:
Resolving the task involves making larger changes than expected, so it is closed and a new task with a larger scope is created

PR #835, opened 2023-06-02
Polygon switch layer

cazapatamar:


src/pages/Search.jsx
```diff
@@ -1441,7 +1455,7 @@ class Search extends Component {
       newState.rasterUrls = [];
       newState.searchType = "definedArea";
       newState.polygon = null;
-      newState.polygonArea = 0;
+      /* newState.polygonArea = 0; */
```
ManuelStardust, line 1458: This line is commented but it is no longer useful

ManuelStardust (changes requested):

src/pages/Search.jsx
cazapatamar: Fixed!

src/pages/search/SearchContext.tsx
```diff
@@ -6,10 +6,8 @@ export interface SearchContextValues {
   geofenceId: string | number;
   searchType: "definedArea" | "drawPolygon";
   polygon: Polygon | null;
-  polygonFolder: string;
-  polygonArea: number;
   setPolygonArea(value: number): void;
-  switchLayer(layer: string): void;
+  switchLayer(layer: string | undefined | null): void;
```
erikasv, line 10: these options are not necessary

ManuelStardust (approved):

erikasv (changes requested):
Sorry I forgot to ask the changes in the comment I left before.

src/pages/search/SearchContext.tsx
cazapatamar: As `switchLayer` has these cases (`undefined` and `null`) i change here the options that function could receive.

cazapatamar: Fixed!

erikasv (approved):

PR #841, opened 2023-07-11
Add dependency and send information to google analytics

ManuelStardust:


erikasv:
You didn't remove the old tag from `index.html`

ManuelStardust:
> You didn't remove the old tag from `index.html`

Done

src/pages/Search.jsx
```diff
@@ -59,12 +60,14 @@ class Search extends Component {
...
   componentDidUpdate() {
-    const { history } = this.props;
+    const { history, location } = this.props;
     history.listen((location, action) => {
       if (location.search === '' && (action === 'PUSH' || action === 'POP')) {
         this.handlerBackButton();
       }
     });
+
+    ReactGA.send({ hitType: "pageview", page: location.pathname + location.search, title: "Consultas" });
```
erikasv, line 70: I don't think this is a good place to put this, maybe send it only if the search string changed, otherwise I think it will be getting a lot more data than the real

ManuelStardust: I put the sending of the page view event when a defined area is selected, in this way the event is not sent multiple times.

erikasv: good call :+1:

erikasv (approved):

PR #851, opened 2023-08-01
Set empty value for layers folder

ManuelStardust:


erikasv:
what is the issue connected to this PR ?

ManuelStardust:
> 
I put it in the same branch of the previous pull reuqest because the error appeared when the actions were executed, however if you consider that it should be a separate issue I can create it

erikasv:
o.O I'm not talking about the branch, but the issue, there is no connected issue (zenhub)

cazapatamar (approved):

PR #858, opened 2023-09-07
Feature/migrate footer to typescript

hacheG:


src/app/layout/Footer.tsx
```diff
@@ -0,0 +1,120 @@
...
+  monitoreo: ["usaid", "geobon", "umed", "temple"],
+};
+
+interface Logo {
+  logosId: string;
+}
+
+const Footer: React.FC<Logo> = ({ logosId }) => (
+  <footer>
+    {logosId && logoSet[logosId as keyof typeof logoSet] ? (
```
erikasv, line 62: This is unnecessarily confusing, the type of logoSet is nameLogo, which you know is an object, so it has keys, so you could ask directly if logosId is a keyof NameLogo.

But we are already saying (with the change suggested in line 57) that logosId must be a key of NameLogo, so I think the right part of the if is unnecesary.

src/app/layout/Footer.tsx
```diff
@@ -0,0 +1,120 @@
...
+      <div className="footerflex">
+        <div>
+          <a href="http://www.humboldt.org.co/es/">
+            <img src={logohumboldt} alt="" />
+          </a>
+        </div>
+        <div className="colaboradores">
+          <h4>Colaboradores</h4>
+          {logoSet[logosId as keyof typeof logoSet].map((name) => {
+            if (!logosData[name as keyof typeof logosData]) return null;
```
erikasv, line 72: In this if, you can (and should) use the variable directly, name comes from a map, that by definition will be a value from a LogoSet property, the whole casting/check is unnecessary.

src/app/layout/Footer.tsx
```diff
@@ -0,0 +1,120 @@
+import React from "react";
+import logohumboldt from "images/logohumboldt.png";
+import logoSiac from "images/logosiac.png";
+import nasa from "images/nasa.png";
+import temple from "images/temple.png";
+import geobon from "images/geobonlogo.png";
+import usaid from "images/usaidlogo.png";
+import umed from "images/umed.png";
+
+interface ImgLogos {
```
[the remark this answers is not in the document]
erikasv: Instead of listing every value and repeating the inner object definition, create a type for the accepted values and change this interface to use it. See [this](https://bobbyhadz.com/blog/typescript-index-signature-parameter-cannot-be-union-type) for a guide

erikasv (changes requested):
@ManuelStardust @cazapatamar you too read this answer to understand more about keyof and typeof: https://stackoverflow.com/questions/55377365/what-does-keyof-typeof-mean-in-typescript/62764510#62764510

src/app/layout/Footer.tsx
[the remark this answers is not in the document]
erikasv: Also I made this example in the playground: https://www.typescriptlang.org/play?#code/C4TwDgpgBA0gogTQMpQLxQEQAsIBtcD2GUAPpgO4EBOuAJhgNwCwAUK6JFACoQDOwaKAG9WUMVADaAawggoASwB2sREgC6ALij8qSgObMWAX1asAxgUX8owPsAAiAQ2COtPa+hEtxmHPiJaGABWjgBujrxmumDAGAA0ouIYlDT0gQAsAEzxrEYMQA

erikasv:
Also connect the zenhub issue

src/app/layout/Footer.tsx
```diff
@@ -0,0 +1,94 @@
...
+  umed: { img: umed, url: "https://udemedellin.edu.co/" },
+};
+
+const logoSet: LogosConfig = {
+  default: ["nasa", "temple", "siac"],
+  monitoreo: ["usaid", "geobon", "umed", "temple"],
+};
+
+interface FooterProps {
+  logosId: "default" | "monitoreo" | null;
```
erikasv, line 35: could you try something like `keyof LogosConfig | null`. I'm not sure it'll work, but would be nice to let it in terms of the keys of LogosConfig instead of the manual list of values

erikasv (approved):

PR #860, opened 2023-09-21
Add percentage to graphs on Landscape

hacheG:


erikasv (changes requested):
- The issue specifically says to calculate the percentage in the backend. So don't calculate it in the chart
- You're missing the other sections in landscape tab, there are only changes for forests
- Use english for variables and functions names

src/pages/search/drawer/landscape/forest/ForestIntegrity.tsx
```diff
@@ -51,6 +51,7 @@ interface FIState {
       id: SCIHFCats;
       label: string;
       value: number;
+      per: number;
```
erikasv, line 54: use the whole word

erikasv (approved):

PR #863, opened 2023-10-09
Add typescript to Header and Layout

hacheG:


src/app/Layout.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Footer from "app/layout/Footer";
+import Header from "app/layout/Header";
+
+interface Names {
+  parent: string | null;
+  child: string | null;
+}
+
+interface LayoutProps {
+  children: any | null;
```
erikasv, line 13: I don't see why null should be allowed.

Also we try to use `any` only when is strictly necessary, in this case children must be components, there is a way to define that type.

src/app/Layout.tsx
```diff
@@ -0,0 +1,36 @@
...
+  moduleName: string;
+  footerLogos: any;
```
erikasv, line 15: footerLogos has a type we defined when migrating Footer to typescript

src/app/Layout.tsx
```diff
@@ -0,0 +1,36 @@
...
+  headerNames: Names;
+  uim: React.ReactNode | null;
```
erikasv, line 17: why is null allowed? if it's an optional property there is a correct syntax for that

src/app/Layout.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import Footer from "app/layout/Footer";
+import Header from "app/layout/Header";
+
+interface Names {
+  parent: string | null;
+  child: string | null;
```
erikasv, line 9: I don't think they can be null. Based on what I see in App.jsx I think they are optional

src/app/layout/Header.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import PropTypes from "prop-types";
```
erikasv, line 2: this isn't used

src/app/layout/Header.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import Menu from "app/layout/header/Menu";
+import Title from "app/layout/header/Title";
+
+interface Names {
+  parent: string | null;
+  child: string | null;
```
erikasv, line 9: dame comment as in Layout

src/app/layout/Header.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+interface HeaderProps {
+  activeModule: string;
+  headerNames: Names;
+  uim: React.ReactNode | null;
```
erikasv, line 15: same comment as in Layout

src/app/layout/Header.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const Header: React.FunctionComponent<HeaderProps> = ({
+  activeModule = "",
```
erikasv, line 19: why does this prop have a default value?

erikasv (changes requested):

src/app/Layout.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { ReactElement } from "react";
+
+import Footer from "app/layout/Footer";
+import Header from "app/layout/Header";
+
+import { LogosConfig } from "../../src/types/layoutTypes";
```
erikasv, line 6: use absolute path (as Footer and Header above)

src/app/layout/Footer.tsx
```diff
@@ -7,16 +7,12 @@ import geobon from "images/geobonlogo.png";
 import usaid from "images/usaidlogo.png";
 import umed from "images/umed.png";
 
-type KEYS = "nasa" | "temple" | "siac" | "geobon" | "geobon" | "usaid" | "umed";
+import { KEYS, LogosConfig } from "../../../src/types/layoutTypes";
```
erikasv, line 10: use absolute path

erikasv (changes requested):

erikasv:
with "absolute path" I mean relative to src

erikasv (approved):

PR #866, opened 2023-10-18
Feature/migrate to typescript menu and title components

hacheG:


src/app/layout/header/Menu.tsx
```diff
@@ -0,0 +1,85 @@
...
+            </li>
+          </Link>
+          { user ? (
+            <Link to="/GEB/Compensaciones" onClick={changeMenuState}>
+              <li>
+                Compensaciones
+              </li>
+            </Link>
+          )
+            : '' }
```
erikasv, line 60: please change this to match the syntax used in Alerts and CBMDashboard links. Instead of using the [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator#syntax) use the [logical AND](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND)

erikasv (changes requested):

src/app/AppContext.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export interface AppContextValue {
+  user: null;
```
erikasv, line 4: user can be null but also can be an object with the form:
```
{
  "id":<int>,
  "username":<string>,
  "name":<string>,
  "company": {
    "id":<int>
  }
}
```

so please add this to the type definition (with correct typescript values)

erikasv (changes requested):

PR #867, opened 2023-10-26
add feature flag for hide portfolio tag

hacheG:


src/pages/search/Drawer.tsx
```diff
@@ -87,10 +95,14 @@ class Drawer extends React.Component<Props, State> {
...
-            { label: "Paisaje", icon: <Paisaje /> },
-            { label: "Especies", icon: <Especies /> },
-            { label: "Portafolios", icon: <Portafolios /> },
+            { label: "Ecosistemas", icon: <Ecosistemas />, disabled: true },
+            { label: "Paisaje", icon: <Paisaje />, disabled: true },
+            { label: "Especies", icon: <Especies />, disabled: true },
+            {
+              label: "Portafolios",
+              icon: <Portafolios />,
+              disabled: this.state.showPortfolios,
```
cazapatamar, line 104: Did you agree with Cesar to hide or disable the Portfolios tab? According to what we discussed yesterday, I understood that Cesar wanted to completely hide the tab.

src/pages/search/shared_components/TabContainer.tsx
```diff
@@ -54,15 +55,18 @@ class TabContainer extends React.Component<Props, State> {
...
+                disabled && (
+                  <Tab
+                    className={`tabs ${tabClasses}`}
+                    label={label}
+                    icon={icon}
+                    key={i}
+                    disabled={disabled}
+                  />
+                )
+            )}
```
cazapatamar, line 69: I think you may be confusing the 'hide' and 'disable' options. If the intention is to hide, then don't use the 'disable' attribute.

cazapatamar (changes requested):

src/pages/search/Drawer.tsx
```diff
@@ -87,10 +95,14 @@ class Drawer extends React.Component<Props, State> {
...
-            { label: "Especies", icon: <Especies /> },
-            { label: "Portafolios", icon: <Portafolios /> },
+            { label: "Ecosistemas", icon: <Ecosistemas />, showTab: true },
+            { label: "Paisaje", icon: <Paisaje />, showTab: true },
+            { label: "Especies", icon: <Especies />, showTab: true },
+            {
+              label: "Portafolios",
+              icon: <Portafolios />,
+              showTab: this.state.showPortfolios,
+            },
```
cazapatamar, line 105: Great, the flag works! But I would like to suggest a change to make the code more readable. Set the property 'showTab' as optional and give it a default value of 'true'. This means that in the Drawer component, you only need to set the 'showTab' attribute in the case of portfolios.

cazapatamar (changes requested):

cazapatamar (approved):

PR #873, opened 2023-11-02
Feature/migrate uim to hooks and typescript

hacheG:


src/app/Uim.tsx
```diff
@@ -0,0 +1,115 @@
...
+   * modal to true, and the others to false.
+   *
+   * @params {String} modal modal name to turn on
+   *
+   * @returns {function}
+   */
+
+  const showModal = (modal: string) => () => {
+    setModals({ ...modals, [modal]: true });
+  };
```
cazapatamar, line 40: What the switch did before was set the state with the name 'modal' (parameter) to true and set the others to false. What you implemented works but doesn't do the same because by using the spread operator, you leave the rest of the states as they are (without changing their values) instead of forcing them to be false

cazapatamar (changes requested):

src/app/Uim.tsx
```diff
@@ -0,0 +1,127 @@
...
+      case "logoutModal":
+        setModals({ loginModal: false, [modal]: true, userModal: false });
+        break;
+      case "userModal":
+        setModals({ loginModal: false, logoutModal: false, [modal]: true });
+        break;
+      default:
+        break;
+    }
+  };
```
cazapatamar, line 51: To avoid using the big switch case, please try using 'showModal' as you had it, but instead of using the spread operator over 'modals,' use 'defaultModalsValues' as you already have the state with default values (false).

Try it and let me know.

(Sorry for being late, we can discuss it on Thursday)

cazapatamar (changes requested):

PR #875, opened 2023-11-10
Feature/app migrate to hooks and typescript

hacheG:


src/App.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+interface LoadComponentTypes {
+  logoSet: keyof LogosConfig | null;
+  name: string;
+  component: React.ReactNode;
+  className: string;
+}
+
+const App: React.FunctionComponent = () => {
+  const [user, setUser] = useState<null>(null);
```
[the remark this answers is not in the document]
erikasv: Also, instead of just null in the type of useState, set the newly defined type as an option

src/App.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  const location = useLocation();
+
+  useEffect(() => {
+    isFlagEnabled("CBMModule").then((value) => setShowCBMDashboard(value));
+  }, []);
+
+  const buildQuery = (queryString: string) => new URLSearchParams(queryString);
+
+  const setUserFunc = (user: null | Response | any): void => setUser(user);
```
erikasv, line 42: I don't think this function is necessary, try removing it and passing `setUser` directly to Uim

src/App.tsx
```diff
@@ -0,0 +1,156 @@
...
+      headerNames: { parent, child },
+    });
+  };
+
+  const loadHome = () =>
+    loadComponent({
+      logoSet: "default",
+      name: "",
+      component: <Home referrer={location.pathname} />,
+      className: "fullgrid",
```
erikasv, line 55: Home didn't have this class, why did you added it?

erikasv (changes requested):

src/App.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const location = useLocation();
+
+  useEffect(() => {
+    isFlagEnabled("CBMModule").then((value) => setShowCBMDashboard(value));
+  }, []);
+
+  const buildQuery = (queryString: string) => new URLSearchParams(queryString);
+
+  const setUserFunc = (user: UserTypes | null) => setUser(user);
```
erikasv, line 44: I did the change to pass directly setUser and it works. This function is unnecessary.

src/types/loginUimProps.ts
```diff
@@ -0,0 +1,14 @@
+export interface CompanyTypes {
```
erikasv, line 1: don't export this type

erikasv (changes requested):

ManuelStardust (approved):

cazapatamar (approved):

PR #881, opened 2023-11-21
Add ConfirmationModal.tsx file

hacheG:


src/components/ConfirmationModal.jsx
```diff
@@ -1,49 +0,0 @@
...
-  styleCustom: PropTypes.string,
-  onContinue: PropTypes.func.isRequired,
-  onCancel: PropTypes.func.isRequired,
-};
-
-ConfirmationModal.defaultProps = {
-  open: false,
-  onClose: null,
-  styleCustom: null,
-};
```
cazapatamar, line 47: By passing to TS, you are missing the required (mandatory) props and default values

hacheG: ok, I changed the props: open to false and styleCustom to null, but the onClose function cannot be sent to null by default. On the other hand, the interface values are required by not having the symbol: ?

cazapatamar: Try using an empty function ()=>{} and read about whether it is possible to use it to replace the null value for onClose function.

Yes, ? is used to make an attribute non-mandatory in the interface. I understand that if you don't use ?, the attribute is mandatory.

hacheG: it's so strange, I changed to () => {} and the Uim.tsx shows me an error ( 102 onClose= ), but biotablero doesn't crash

cazapatamar: I just make the change using () => {} as a default value for onClose and it works for me

cazapatamar (approved):

PR #889, opened 2023-12-11
Feature/information migrate to hooks and typescript

hacheG:


src/pages/home/Information.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const Information: React.FC<InformationTypes> = ({ activeModule }) => {
+  const [activeItem, setActiveItem] = useState<string>("queEs");
+
+  type Keys = "queEs" | "porque" | "quienProduce" | "queEncuentras";
+
+  interface ContentInfoTypes {
+    [key: string]: BasicTitleTypes;
+  }
```
cazapatamar, line 23: Why did you put this type and interface definition here?. I think it is more redeable if you leave all interfaces and types at the top (above or below InformationTypes). What do you think?

hacheG: yes you have reason, i' ll change

cazapatamar (approved):

PR #903, opened 2024-01-23
Feature/migrate nivo component dots graph

hacheG:
para probar el código es necesario comentar el setHeaderNames en el componente App.tsx
image


src/pages/compensation/drawer/graphLoader/DotsScatterPlot.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+export const DotsScatterPlot: React.FC<ScatterProps> = ({
+  activeBiome,
+  dataJSON,
+  labelX,
+  labelY,
+  colors,
+  elementOnClick,
+}) => {
+  dataList = dataJSON.map((axis) => {
```
erikasv, line 56: axis seems like a bad name for the elements of dataJSON

src/pages/compensation/drawer/graphLoader/DotsScatterPlot.tsx
```diff
@@ -0,0 +1,196 @@
...
+      }}
+      colors={(obj) => {
+        return getColor(String(obj.serieId));
+      }}
+      blendMode="multiply"
+      enableGridX={true}
+      enableGridY={true}
+      nodeSize={(obj) => {
+        //console.log("Z:",obj.data);
+
```
erikasv, line 148: delete these lines

src/pages/compensation/drawer/graphLoader/DotsScatterPlot.tsx
```diff
@@ -0,0 +1,196 @@
...
+    };
+    return colorsObj[serieId] ?? colors[3];
+  };
+  getSize = (serieId, xValue, biome) => {
+    dataBiome = biome;
+    let x = parseFloat(xValue);
+    sizes = {
+      High: x >= 80 ? 30 : x + 4,
+      Medium: x === 0 ? 15 : x >= 80 ? 30 : x + 4,
+      Low: x === 0 ? 15 : x + 4,
```
erikasv, line 105: I haven't checked the functionality, but these seem a little odd

erikasv (changes requested):

src/pages/compensation/drawer/graphLoader/DotsScatterPlot.tsx
erikasv: @hacheG leave a comment in this function informing that this used to be a d3 scale, and include this link: https://github.com/PEM-Humboldt/biotablero-frontend/blob/34052510ee224c03439edd4a8b531e4929246272/src/pages/compensation/drawer/graphLoader/DotsGraph.jsx#L28

src/pages/compensation/drawer/graphLoader/DotsScatterPlot.tsx
```diff
@@ -0,0 +1,207 @@
...
+            affectValue.affected_natural === ""
+              ? "0"
+              : affectValue.affected_natural,
+          biome: affectValue.name,
+        },
+      ],
+    };
+  });
+  const getColor = (serieId: string): string => {
+    if (dataBiome === biomeComparator && serieId === "High") {
```
erikasv, line 81: `dataBiome` is doing the same as `activeBiome` prop, and since it's used by the parents of this component it's better handled as prop. Just delete this state and use `activeBiome` again.

The issue is using `biomeComparator` as reference for the selected biome (you just changed its name).

I figured out why it works and it's because nivo exectues, for each point, the size function first and then the color function, so `biomeComparator` changes value just before it's used in the color function. Still, you need to change this implementation because is using a collateral effect rather than an intended workflow.

As mentioned in a previous meeting you must change the `id` assigned for the `dataList` items making it enough to identify the point biome and category. A possible solution can be concatenating those values and splitting them in the colors function.

erikasv (changes requested):

src/pages/compensation/drawer/graphLoader/DotsScatterPlot.tsx
erikasv: Finally a custom component was defined, this leaves a problem with the tooltip, but it was possible to keep the previous implementation of the stroke for the selected node

erikasv (approved):

PR #904, opened 2024-01-24
Feature/clean map on polygon

hacheG:


src/pages/search/Selector.jsx
```diff
@@ -33,24 +35,27 @@ const Selector = (props) => {
...
     },
   ];
 
   const onChange = (level, tabId) => {
-    if (tabId === 'draw-polygon') {
+    if (tabId === "draw-polygon") {
+      handlers.areaListChange();
       handlers.polygonChange();
     } else {
       handlers.areaListChange();
```
cazapatamar, line 61: I don't like this approach. 

1) you are calling the function ´areaListChange()´ when de condition is true and false, so in these cases you can simply put it out of the condition.

2) I like better try to inactivate all layers in Search instead of here. Check shutOffLayer funtion and try to use it inside polygonChange handler.

src/pages/search/selector/SearchAreas.jsx
cazapatamar, line 1: What changes did you make in this file? I couldn't identify them

cazapatamar (changes requested):

src/pages/search/Selector.jsx
hacheG: reorganize the code because the "shutOffLayer " function is not executed when clicking the "Área de consulta"  button

erikasv (changes requested):
Undo the formatting changes, it's impossible to identify the relevant changes

hacheG:
It was very difficult to review the changes because to a prettier update

PR #905, opened 2024-01-24
Fix HeaderNames bug

ManuelStardust:


cazapatamar:
The big is solved! I just wonder whether the console message is related to this bug:

imagen

Steps to reproduce:

1. Log in into aplication
2. Enter to Search module
3. Select any geofence
4. Back to home
5. Enter to compensation module

cazapatamar (approved):

PR #906, opened 2024-02-09
move areaListChange function

hacheG:
## 📝 Associated issues
#900 

## 🤔 Considerations
I closed the previous PR because I made a prettier update

### DIFF
--- a/src/pages/Search.jsx
+++ b/src/pages/Search.jsx
@@ -1600,6 +1600,7 @@ class Search extends Component {
                     geofenceChange: this.innerElementChange,
                     polygonChange: () => {
                       this.setState({ drawPolygonEnabled: true });
+                      this.shutOffLayer()
                     },
                   }}
                   description={Description()}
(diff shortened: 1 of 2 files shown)


cazapatamar (commented):
What i meant in the comment i left in the last PR already closed, it was related to avoid calling areaListChange(); from Selector. In Search I think it may works simply calling shutOfLayers inside the definition of polygonChange (prop of Selector). What do u think? @hacheG @erikasv

erikasv:
> What I meant in the comment I left on the last closed PR was related to avoiding calling areaListChange(); from the Selector. In the Search component, I think it may work simply by calling shutOffLayers inside the definition of polygonChange (a prop of Selector). What do you think? @hacheG @erikasv

I actually don't know if this change creates a side effect by not calling ´handlers.areaListChange();´ when drawing a polygon. I'd have to test the functionality, if I see something odd I'll let you know

PR #931, opened 2025-01-10
FLP layers won't show if component is unmounted

erikasv:
## 🛠️ Changes
Fix the issue with forest loss persistence layer being shown after the section is closed

## 📝 Associated issues
#927


erikasv:
will change base branch to avoid conflicts

PR #954, opened 2025-02-15
Migrate Timeline HF Layers

ManuelStardust:
## 🛠️ Changes
Migrating layer handling for timeline and SE human footprint component:
- Add controller for HF Timeline component
- Migrate layers request from RestAPI to BackendAPI
- Adjust logic in HF Timeline component according to new Search component

## 📝 Associated issues
#949


erikasv:
@ManuelStardust alguna razón por la que esta rama no se haya eliminado?

